Issue #262, opened 2014-07-25
Conditions for Guards


davidwindell:
@jmleroux I need an OR condition on the RoutePermissionGuard - did you have any time to implement this?

PR #268, opened 2014-08-26
Conditions for Guards

davidwindell:
See #262.

So far this only applies to the RoutePermissionGuard as a proof of concept.

It's a shame that this guard specifically was created with an AND default whereas an OR would have made more sense (like the RouteGuard).

### DIFF
--- a/docs/04. Guards.md
+++ b/docs/04. Guards.md
@@ -177,7 +177,7 @@ return [
         'guards' => [
             'ZfcRbac\Guard\RoutePermissionsGuard' => [
                 'admin*' => ['admin'],
-                'post/manage'   =>
+                'post/manage' => ['post.update', 'post.delete']
             ]
         ]
     ]
--- a/docs/04. Guards.md
+++ b/docs/04. Guards.md
@@ -177,17 +177,34 @@ return [
         'guards' => [
             'ZfcRbac\Guard\RoutePermissionsGuard' => [
                 'admin*' => ['admin'],
-                'post/manage'   => ['post.update', 'post.delete']
+                'post/manage'   =>
             ]
         ]
     ]
 ];
 ```
 
-> All permissions in a rule must be matched (it is an AND condition).
+> By default, all permissions in a rule must be matched (an AND condition).
 
 In the previous example, one must have ```post.update``` **AND** ```post.delete``` permissions
-to access the ```post/manage``` route.
+to access the ```post/manage``` route. You can also specify an OR condition like so:
+
+```php
+use ZfcRbac\Guard\GuardInterface;
+
+return [
+    'zfc_rbac' => [
+        'guards' => [
+            'ZfcRbac\Guard\RoutePermissionsGuard' => [
+                'post/manage'   => [
+                    'permissions' => ['post.update', 'post.delete'],
+                    'condition'   => GuardInterface::CONDITION_OR
+                ]
+            ]
+        ]
+    ]
+];
+```
 
 > Permissions are linked to roles, not to users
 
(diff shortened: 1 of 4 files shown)


davidwindell:
@bakura10 Just noticed this one - can you get it merged now?

docs/04. Guards.md
```diff
@@ -177,17 +177,34 @@ return [
         'guards' => [
             'ZfcRbac\Guard\RoutePermissionsGuard' => [
                 'admin*' => ['admin'],
-                'post/manage'   => ['post.update', 'post.delete']
+                'post/manage'   =>
```
bakura10: Doc is wrong here :p.

src/ZfcRbac/Guard/RoutePermissionsGuard.php
```diff
@@ -101,12 +100,30 @@ public function isGranted(MvcEvent $event)
...
+            ) {
+                return true;
+            } elseif ($condition === GuardInterface::CONDITION_AND
+                      && !$this->authorizationService->isGranted($permission)
+            ) {
                 return false;
             }
         }
 
+        if ($condition === GuardInterface::CONDITION_OR) {
```
bakura10: Hmmm I'm trying to read the code and I'm not sure to understand this.

bakura10:
Hi,

Sorry about this, we should definitely start to merge it :p. @danizord could you have your own quick review please ?

davidwindell:
Great - we have been using in production for some time now. Works grand.

bakura10:
Ok, I'm going to merge than Monday if nothing complain about it. Can you remind me if I forget?

tests/ZfcRbacTest/Guard/RoutePermissionsGuardTest.php
```diff
@@ -326,6 +326,26 @@ public function routeDataProvider()
                 'isGranted'           => true,
                 'policy'              => GuardInterface::POLICY_DENY
             ],
+            [
+                'rules'               => ['route' => [
```
danizord: No need to align when the value is array

src/ZfcRbac/Guard/RoutePermissionsGuard.php
```diff
@@ -101,12 +100,30 @@ public function isGranted(MvcEvent $event)
             return true;
         }
 
-        foreach ($allowedPermissions as $permission) {
-            if (!$this->authorizationService->isGranted($permission)) {
+        $permissions = isset($allowedPermissions['permissions'])
```
danizord: Should probably happen before wildcard check

danizord:
I think you can simplify the logic to something like that:
```php
if (GuardInterface::CONDITION_AND === $condition) {
    foreach ($permissions as $permission) {
        if (!$this->authorizationService->isGranted($permission)) {
            return false;
        }
    }

    return true;
}    

if (GuardInterface::CONDITION_OR === $condition) {
    foreach ($permissions as $permission) {
        if ($this->authorizationService->isGranted($permission)) {
            return true;
        }
    }

    return false;
}    

throw new InvalidArgumentException(sprintf(
    'Condition must be either "AND" or "OR", %s given',
    is_object($condition) ? get_class($condition) : gettype($condition)
));
```

bakura10:
I agree with @danizord change, more efficient and way easier to read.

bakura10:
Thanks!

Just a few more CS fixes to make travis happy: https://travis-ci.org/ZF-Commons/zfc-rbac/jobs/56477072

davidwindell:
:+1:

bakura10:
Thanks! I've tagged ZfcRbac as 2.5.0 :).

Hopefully someone will be able to backport this feature for other guards !

PR #289, opened 2015-03-12
Replace ObjectManager with Selectable

fabiocarneiro:
```findBy(['property' => []])``` is a feature available in ```EntityManager``` only. Since you're typehinting ```ObjectManager``` instead of ```EntityManager```, is expected that this will work if you provide a MongoDB ODM ```DocumentManager```, which unfortunately does not happen given the lack of support for values in arrays.

That said, i propose to use ```Selectable``` instead of ```ObjectManager```, so it will be possible to use even a ```Role``` ```Collection``` to retrieve the Roles.

This might seem very strange not to consider this a BC Break, but since you're typehinting ```ObjectManager``` the expected behavior is already to work with any manager, and if this is not happening, i would consider this more a bugfix than a bc break. Also, the only possible working setting is ```EntityManager```, which is a ```Selectable``` instance, so it will be perfectly compatible with any previous implementation.

### DIFF
--- a/src/ZfcRbac/Role/ObjectRepositoryRoleProvider.php
+++ b/src/ZfcRbac/Role/ObjectRepositoryRoleProvider.php
@@ -18,11 +18,12 @@
 
 namespace ZfcRbac\Role;
 
-use Doctrine\Common\Persistence\ObjectRepository;
+use Doctrine\Common\Collections\Criteria;
+use Doctrine\Common\Collections\Selectable;
 use ZfcRbac\Exception\RoleNotFoundException;
 
 /**
- * Role provider that uses Doctrine object repository to fetch roles
+ * Role provider that uses Doctrine selectable to fetch roles
  *
  * @author  Michaël Gallego <mic.gallego@gmail.com>
  * @licence MIT
@@ -30,9 +31,9 @@ use ZfcRbac\Exception\RoleNotFoundException;
 class ObjectRepositoryRoleProvider implements RoleProviderInterface
 {
     /**
-     * @var ObjectRepository
+     * @var Selectable
      */
-    private $objectRepository;
+    private $selectable;
 
     /**
      * @var string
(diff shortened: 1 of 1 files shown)


danizord:
@fabiocarneiro The fact that `$repository->findBy(['field_name' => ['a', 'b', 'c']])` fails is actually a limitation of the `Doctrine\ODM\MongoDB\DocumentRepository` implementation, not a ZfcRbac issue at all.

Having said that, I'd suggest you to keep `ObjectRepositoryRoleProvider` untouched and create a new `SelectableRoleProvider` that works with `Selectable` instances. Just to keep things backwards compatible.

What do you think?

fabiocarneiro:
@danizord With that the way it is, this is only compatible with EntityManager. There is nothing forcing any implementation of ObjectManager to accept that array of values, so you're relying in something that you can't rely. There is no contract for that. 

On the other side, Selectable is a perfect contract for that and will work no matter which objectmanager you're using or even a collection.

It will still be Backwards Compatible because the only possible implementation right now is EntityManager and EntityManager is instanceof Selectable.

bakura10:
You are confusing the ObjectManager and ObjectRepository interfaces here. ObjectRepository definitely have a findBy: https://github.com/doctrine/common/blob/master/lib/Doctrine/Common/Persistence/ObjectRepository.php#L64

So as @danizord said, if there is a problem in that, this must be solved on DoctrineODM, not here. We are respecting the contract ;).

fabiocarneiro:
@bakura10 There is no contract for the array as value. We agreed that this is not a MongoDB ODM issue, but zfc-rbac, since you're relying in something you shouldn't. We've discussed it and I'll try another approach.

bakura10:
hmmmm I see... Actually I'm not really against changing this against criteria. But rather than modifying it, I'd prefer if you could actually create a new SelectableRoleProvider that would extend ObjectRepository.

The main problem I see is that getRepository in ObjectManager returns a ObjectRepository, not a Selectable.

PR #291, opened 2015-04-14
collectPermissions check if getPermissions method exists

51systems:
If you are using a HierarchicalRole backed by doctrine where the children are lazy loaded, the collectPermissions method fails on a child role. This is because the child role is an instance of a doctrine proxy object, and does not have a permissions property to reflect.

Instead of immediately falling back to reflection, we can check if a getPermissions method exists on the role object and call that first.

Still not a very elegant solution, but it makes it a bit more robust.

### DIFF
--- a/tests/ZfcRbacTest/Collector/RbacCollectorTest.php
+++ b/tests/ZfcRbacTest/Collector/RbacCollectorTest.php
@@ -29,6 +29,8 @@ use ZfcRbac\Role\InMemoryRoleProvider;
 use ZfcRbac\Role\RoleProviderInterface;
 use ZfcRbac\Service\RoleService;
 use Rbac\Traversal\Strategy\RecursiveRoleIteratorStrategy;
+use ZfcRbacTest\Asset\MockRoleWithPermissionMethod;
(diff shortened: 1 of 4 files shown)


bakura10:
Mmhh considering this is only done on the debugging tool I'm ok for that... However you'll need to write a small test to ensure we don't have regression in the future.

tests/ZfcRbacTest/Collector/RbacCollectorTest.php
```diff
@@ -176,4 +178,126 @@ public function testCanCollect()
...
+    {
+        return 'role-with-permission-property';
+    }
+    public function hasPermission($permission)
+    {
+        return false;
+    }
+}
+
+class MockRoleWithPermissionMethod implements RoleInterface
```
bakura10: You should create those two classes in the asset folder: https://github.com/ZF-Commons/zfc-rbac/tree/master/tests/ZfcRbacTest/Asset :)

bakura10:
Thanks. I will tag this today.

bakura10:
Done, 2.5.1 has been tagged with this fix.

Thanks a lot ! :)

PR #296, opened 2015-05-12
Fix a bug with zend developer tools ZDT

belgattitude:
Latest ZDT update comes breaks ability to profile zfcrbac. Their commit can be seen :

https://github.com/zendframework/ZendDeveloperTools/commit/d3432681aa32177a741ad23604a40af9ad454acb#commitcomment-11090423

After a quick look, their fix seems relevant as relying on serializable behaviour is not very clear. 

This PR seems to fix the issue, by correctly returning collection as an array.

### DIFF
--- a/src/ZfcRbac/Collector/RbacCollector.php
+++ b/src/ZfcRbac/Collector/RbacCollector.php
@@ -46,11 +46,6 @@ class RbacCollector implements CollectorInterface, Serializable
     /**
      * @var array
      */
-    protected $collection = [];
-
-    /**
-     * @var array
-     */
     protected $collectedGuards = [];
 
     /**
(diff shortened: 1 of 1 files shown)


bakura10:
Hi,

Thank you for this. Tests are failing though, they need to be updated too ?

belgattitude:
@bakura10 

not so sure... I should have been more attentive to unserialize method. the collection should be restored to individual attributes, like

```php
    /**
     * @return array|string[]
     */
    public function getCollection()
    {
        return [
            'guards'      => $this->collectedGuards,
            'roles'       => $this->collectedRoles,
            'permissions' => $this->collectedPermissions,
            'options'     => $this->collectedOptions
        ];
    }
    /**
     * {@inheritDoc}
     */
    public function serialize()
    {
        return serialize($this->getCollection());
    }
    /**
     * {@inheritDoc}
     */
    public function unserialize($serialized)
    {
        $collection = unserialize($serialized);
        $this->collectedGuards = $collection['options'];
        $this->collectedRoles  = $collection['roles'];
        $this->collectedPermissions =  $collection['permissions'];
        $this->collectedOptions =  = $collection['guards'];
    }
```

Looks a bit weird to me, but should do the trick. Have you an idea ?

belgattitude:
My first PR and removed my fork after making it, see this one instead :+1: 
https://github.com/ZF-Commons/zfc-rbac/pull/297

PR #298, opened 2015-05-29
Add support for setter injection of IdentityProviderInterface and RoleeProviderInterface instances

sebastianbergmann:
This solves problems such as the one described at http://stackoverflow.com/questions/29340127/zend-framework-2-abstractcontrollertestcase-and-zfcuser-zfcrbac

### DIFF
--- a/src/ZfcRbac/Service/RoleService.php
+++ b/src/ZfcRbac/Service/RoleService.php
@@ -74,6 +74,26 @@ class RoleService
     }
 
     /**
+     * Set the identity provider
+     *
+     * @param IdentityProviderInterface $identityProvider
+     */
+    public function setIdentityProvider(IdentityProviderInterface $identityProvider)
+    {
+        $this->identityProvider = $identityProvider;
+    }
+
+    /**
+     * Set the identity provider
+     *
+     * @param RoleProviderInterface $roleProvider
+     */
+    public function setRoleProvider(RoleProviderInterface $roleProvider)
+    {
+        $this->roleProvider = $roleProvider;
+    }
+
+    /**
      * Set the guest role
      *
      * @param  string $guestRole


bakura10:
Why is this needed? This is not supposed to be changed, why can't you simply mock and inject them on constructor?

src/ZfcRbac/Service/RoleService.php
```diff
@@ -74,6 +74,26 @@ public function __construct(
...
+     *
+     * @param IdentityProviderInterface $identityProvider
+     */
+    public function setIdentityProvider(IdentityProviderInterface $identityProvider)
+    {
+        $this->identityProvider = $identityProvider;
+    }
+
+    /**
+     * Set the identity provider
```
sebastianbergmann: This should of course be "Set the role provider".

sebastianbergmann:
@bakura10 Thank you for your time to look at this. This is required because in the context of a test that uses Zend Framework 2's `AbstractHttpControllerTestCase` the guards are already registered by the time we can inject mocks into the service locator. Using the setter injection added by this patch we can replace the original instance with out mock which we cannot register otherwise.

bakura10:
Okay, I definitely think that mocking of controller should be done easier in ZF2, but given the current limitation, change sounds sane. I'll merge and tag.

bakura10:
Done. ZfcRbac has been tagged to 2.5.2

sebastianbergmann:
Thanks!

clemf:
@sebastianbergmann Is it possible to give an example of how to use this patch? I am struggling with this exact issue but I don't understand how these setters help.

sebastianbergmann:
@clemf Sorry, I don't remember the details.

Issue #258, opened 2014-07-02
Multiple roles given in context


bakura10:
Hi everyone,

Just a question to know how you guys/girls would solve this problem (I'd like to add that to the cookbook, eventually).

Imagine a project where each user has a given role (most of the time, "member"). In this system, users can create project where they have a bunch of permissions related to their project (like "project.create", "project.delete", "project.add_foo").

Now, you have the ability to invite other users to this project. Those users are users of the system, so they are also member, and they also have the "project.create", "project.delete"... permissions). But when I invite a user, I could give him a specific role _for this project_, so that he is read-only for instance, only on this project.

How would you tackle this problem?

ping @arekkas @danizord

zionsg:
Hi, might the following 2 solutions work? Solution 1 involves passing the project name as $context to getRoles() while the more complicated Solution 2 involves prepending the project name to roles and permissions.

First of all, let's assume 4 tables in the database: role, user, project, map_project_user.
Each user is assigned different roles for each project. A member can do everything but a guest is readonly.
```
role: member, guest
user: Alice, Bob
project: Gamma, Delta

map_project_user
|:-------:|:-----:|:------:|
| project |  user |  role  |
|:-------:|:-----:|:------:|
|  Gamma  | Alice | member |
|  Delta  |  Bob  | member |
|  Delta  | Alice | guest  |
|:-------:|:-----:|:------:|
```

## Solution 1

1) Modify ZfcRbac\Identity\IdentityInterface::getRoles() to take in an optional param `$context`.
```
public function getRoles($context = null);
```

2) The authenticated identity (likely populated from the database) will store its roles for each project and return only the relevant roles given the project name.
```
namespace App\Identity;

use ZfcRbac\Identity\IdentityInterface;

class Identity implements IdentityInterface
{
    public function getRoles($context = null)
    {
        /*
          In this case, $context will be the project name.
          Using Alice as an example, if $context is 'gamma', return ['member'].
          If $context is 'delta', return ['guest'].
          What is safe to return if $context is null?
        */
    }
}
```

3) Modify `ZfcRbac\Service\AuthorizationService::isGranted()` and pass the context to `getIdentityRoles()`.
```
public function isGranted($permission, $context = null)
{
    $roles = $this->roleService->getIdentityRoles($context);

    /* no change to rest of code */
}
```

4) Modify `ZfcRbac\Service\RoleService::getIdentityRoles()` to take in an optional $context param and pass it to `getRoles()`.
[...]

Issue #241, opened 2014-06-15
[RFC] [3.0] Context Aware RBAC


arekkas:
## The problem

RBAC is a great system to provide authorization utilities to a software. When dealing with multiple blended systems (shared multi tenancy) RBAC fails to comply. Let's take an example:

1. UserA is an editor on BlogA
* UserA want's to create a new post on blogA -> success
* UserA want's to create a new blog post blogB -> failure
* UserB grants UserA the editor role on blogB
* UserA want's to create a new blog post blogB -> success

This is  currently impossible to do, because all we can do is:

```php
$this->isGranted('post.create');
```

Sure, there are assertions, but do they any good?

```php
public function assert(AuthroizationService $service, $context = NULL)
{
    // We can check for permissions or ownership here, but there is no way to say if UserA has an editor role on blogB
}
```

### pRBAC

https://github.com/dimagi/django-prbac/blob/master/docs/tutorial.rst#id1
https://github.com/dimagi/django-prbac/blob/master/django_prbac/models.py

I have not fully understood that implementation yet, because they do not use classic terms like role and permission but rather role, grant and group. For example, this is how you achieve a simple RBAC:

```
kenn = Role.objects.create(name='kenn')
may_view_reports = Role.create(name='may_view_reports', friendly_name='May view reports')
Grant.objects.create(from_role=kenn , to_role=may_view_reports)
kenn.has_privilege(may_view_reports)
```

Which doesn't make a lot of sence to me at first glance. PRBAC looks like this:

```
kenn = Role.objects.create(name='kenn')
may_view_reports = Role.create(name='may_view_reports', friendly_name='May view reports')
Grant.objects.create(from_role=kenn , to_role=may_view_reports, assignment={'report_name': 'active_users'})
kenn.has_privilege(may_view_report.instantiate({'report_name': 'active_users'}) // true
kenn.has_privilege(may_view_report.instantiate({'report_name': 'something_else'}) // false
```

### Context-aware (or parametrized) Permissions

![rbac1](https://cloud.githubusercontent.com/assets/3372410/3280791/6bd8e618-f47e-11e3-9374-2d197e7ce7a3.png)

```php
$this->isGranted('post.create', $parameterProvider);
```

```php
public function AuthorizationService::isGranted($permissionGroup, ParameterProvider $parameterProvider)
{
    $roles = $this->roleService->getIdentityRoles();
    return $this->carbac->isGranted($roles, $permissionGroup, $parameterProvider);
}

[...]

PR #299, opened 2015-06-05
Allow optional context for getRoles

zionsg:
This allows the identity to return different roles for different contexts. A scenario would be a user having different roles for different projects and the project name is passed in as the context. Implements Solution 1 in https://github.com/ZF-Commons/zfc-rbac/issues/258#issuecomment-104590211

### DIFF
--- a/src/ZfcRbac/Identity/IdentityInterface.php
+++ b/src/ZfcRbac/Identity/IdentityInterface.php
@@ -29,7 +29,8 @@ interface IdentityInterface
     /**
      * Get the list of roles of this identity
      *
+     * @param  string $context Optional context for determining roles returned, eg. project name as context
      * @return string[]|\Rbac\Role\RoleInterface[]
      */
-    public function getRoles();
+    public function getRoles($context = null);
 }
--- a/src/ZfcRbac/Service/AuthorizationService.php
+++ b/src/ZfcRbac/Service/AuthorizationService.php
@@ -121,7 +121,7 @@ class AuthorizationService implements AuthorizationServiceInterface
      */
     public function isGranted($permission, $context = null)
     {
-        $roles = $this->roleService->getIdentityRoles();
+        $roles = $this->roleService->getIdentityRoles($context);
 
         if (empty($roles)) {
             return false;
--- a/src/ZfcRbac/Service/RoleService.php
+++ b/src/ZfcRbac/Service/RoleService.php
@@ -107,10 +107,11 @@ class RoleService
     /**
      * Get the identity roles from the current identity, applying some more logic
      *
+     * @param  string $context Optional context for determining roles returned, eg. project name as context
      * @return RoleInterface[]
      * @throws Exception\RuntimeException
      */
-    public function getIdentityRoles()
+    public function getIdentityRoles($context = null)
     {
         if (!$identity = $this->getIdentity()) {
             return $this->convertRoles([$this->guestRole]);
@@ -123,7 +124,7 @@ class RoleService
             ));
         }
 
-        return $this->convertRoles($identity->getRoles());
+        return $this->convertRoles($identity->getRoles($context));
     }
 
     /**


src/ZfcRbac/Identity/IdentityInterface.php
```diff
@@ -29,7 +29,8 @@
     /**
      * Get the list of roles of this identity
      *
+     * @param  string $context Optional context for determining roles returned, eg. project name as context
      * @return string[]|\Rbac\Role\RoleInterface[]
      */
-    public function getRoles();
+    public function getRoles($context = null);
```
danizord: Keep optional arguments out of the contract.

bakura10:
I like the idea of being able to pass context, however don't we have a BC here?

src/ZfcRbac/Identity/IdentityInterface.php
bakura10: Why not? It at least enforces that getRoles can receive an optional context.

zionsg:
Yes, it's a BC especially since projects would implement IdentityInterface directly, given that there is no AbstractIdentity to extend from. I did try out Solution 2 in https://github.com/ZF-Commons/zfc-rbac/issues/258#issuecomment-104590211 but it did not work out. The defining of permissions per role for the role provider uses simple permissions without the project name being prepended - there seemed no way to prepend the project name in the whole process. Either I missed out something or it was getting too contrived.

I am using this fork in a project right now and it was easier to setup. The User entity implements the IdentityInterface and is populated with the team roles when fetched by the mapper during authentication. Roles are divided into site roles and project roles.

```
    public function getRoles($context = null)
    {
        if (!$context) {
            return $this->getSiteRoleName();
        }

        return isset($this->teamRoles[$context]) ? $this->teamRoles[$context] : array();
    }
```

bakura10:
Ok, as I told you I like the idea but I'd prefer not doing a BC here. There is high chance that Rbac package (the one that ZfcRbac uses) will be merged to ZF, so I want ZfcRbac 3 to rely on ZF package. I'd prefer not to BC now if I need to release another major version in 2 months.

Regarding the issue itself this is actually a limitation of the rbac model itself. For such cases we would need to use more advanced authorization model like CARBAC but it's definitely harder so none of us had the time to do it.

But your fix provide an easy way to make rbac a bit more flexible. 

Envoyé de mon iPhone

zionsg:
Ya, understood the BC implications. Nevertheless it has been a good learning exercise for me which got me a temporary solution for my current project. Learnt more from https://github.com/ZF-Commons/zfc-rbac/issues/241 after you mentioned CARBAC. Thanks for your feedback and advice (and danizord's too)!

danizord:
Well, maybe I'm missing something, but I don't see any BC breaks there. You are only adding an optional argument to the interface, current implementations will keep working AFAIK.

Anyway I think this logic should happen in "role provider", not in the identity object. But it's currently impossible because ZfcRbac calls `getRoles()` directly in the identity object. The good news is that I'm planning to propose a change on this behavior for 3.0, so that we can write our custom role providers with specific domain logic.

@bakura10 I'd say it is mergeable for 2.x, but the bad thing is that we are adding a new behavior that will likely change in 3.x. Leaving that decision for you :P

bakura10:
Oh that would be awesome @danizord to be able to write custom role providers.

I'd prefer to not merge that for now (but I'll keep that open for reference), and we'll see the evolution of both Rbac and ZfcRbac component.

Btw any news @Danizord for the Rbac stuff on ZF repo? You still need to convince Matthew and make PR to the branch for removing the non-generator stuff ^^.

danizord:
@bakura10 I was procrastinating this one tbh, haha :P but tonight I will start writing.

imonteiro:
I'm very interested in this topic. In my application the roles depend of which project (aka context) is selected, but I still haven't discovered how I should solve.

@zionsg Can you share a complete example?

Thanks in advance.

zionsg:
@imonteiro I posted 2 possible solutions earlier on, of which I preferred the first. You can read it at https://github.com/ZF-Commons/zfc-rbac/issues/258#issuecomment-104590211

PR #301, opened 2015-07-10
Register guards only for http requests

thestanislav:
Guards should not be registered for CLI applications

### DIFF
--- a/src/ZfcRbac/Module.php
+++ b/src/ZfcRbac/Module.php
@@ -21,6 +21,7 @@ namespace ZfcRbac;
 use Zend\EventManager\EventInterface;
 use Zend\ModuleManager\Feature\BootstrapListenerInterface;
 use Zend\ModuleManager\Feature\ConfigProviderInterface;
+use Zend\Http\PhpEnvironment\Request as HttpRequest;
 
 /**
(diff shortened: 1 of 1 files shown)


danizord:
Not sure if this is mergeable, I think ZfcRbac guards are also usable in CLI env.

Ping @bakura10

thestanislav:
But right now if any guard is set, for any cli applications we have "Don't have permission" error

danizord:
@thestanislav You can solve this by authenticating some identity (e.g. some worker) when running the application from CLI or overriding `ZfcRbac\Guards` service factory to return an empty array.

If I merge this one, I could break some applications that relies on ZfcRbac to protect their CLI.

bakura10:
This is definitely a BC break.

As danizord say, one option would be to authenticate your CLI requests. If you are using oauth token for instance, you could pass the key as part of your CLI requests to authenticate it.

As a side note, this is a reason why I prefer not to use guard, and only protect my controller directly. If you use guard or protect services, you run into this problem.

If you protect your controllers (which are the entry point of an API most of the time), then management become easier (as you'll have different controllers for console routes, where you can remove the check).

I'm therefore closing for now.

PR #305, opened 2015-07-27
[WIP] permissions list function

Orkin:
@bakura10 what do you think about this ?
Need more test etc ... It's just a POC

Some typo are incorrect getPermissions return string[] in phpdoc but in fact it's Permission[]

### DIFF
--- a/config/module.config.php
+++ b/config/module.config.php
@@ -35,6 +35,7 @@ return [
             'ZfcRbac\Role\RoleProviderPluginManager'          => 'ZfcRbac\Factory\RoleProviderPluginManagerFactory',
             'ZfcRbac\Service\AuthorizationService'            => 'ZfcRbac\Factory\AuthorizationServiceFactory',
             'ZfcRbac\Service\RoleService'                     => 'ZfcRbac\Factory\RoleServiceFactory',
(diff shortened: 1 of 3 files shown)


danizord:
What's the use case?

Orkin:
For retrieve the list of permissions. We can only get role but no permissions. When we use the view helper we don't have to call this method but when you have an API, you need te retrieve permissions on client side and for this an endpoint will be useful.

danizord:
@Orkin mhmm, but this is out of the scope of this library. I think it fits your own application service layer, not ZfcRbac.

Orkin:
I have implemented in my application. It's just sharing in case it had it up here. If this is not the case no problem;).

danizord:
@Orkin thanks for your contribution, but I really think it does not fit in this library =/

@bakura10 what do you think?

bakura10:
You're right, I'd prefer not to complexicate this even more by introducing a new service. While "RoleInterface" does not have a "getPermissions", the concrete implementation has one (https://github.com/zf-fr/rbac/blob/master/src/Rbac/Role/Role.php#L49).

You could therefore add the method "getPermissions" right into your Identity object, as you can have access to all roles and, in turn, all permissions. I think the logic better suit there and avoid introducing a new service.

Thanks anyway ;).

Issue #307, opened 2015-09-22
Embracing middlewares


bakura10:
Hi everyone,

In overall, this package has been very stable over the last few months, and I'd like to think about the future of it.

With the increase of API, and tools like Zend\Expressive, it may be interesting to simplify this package. I'd like to support two different branches: 2.x that would be based as a ZF2 module, and a 3.x branch, that would remove dependency completely toward ZF, and embrace middlewares and an API first design.

Here are the various things I'm thinking, let me know if you have any suggestion.

## Removing guards

Guards were nice, but to be honest, they were a bit problematic. They were tied very strongly to the ZF2 router / ZF2 MVC model, and while useful to "guard" a bunch of route (/admin/*), the advent of middlewares allow to create much better solution to this issue.

Also, guards introduced potential security issue, as some methods could be called outside the context of a route/controller, hence bypassing the guard.

This also made testing harder, because, well... testing taht kind of things is hard because it's 100% config.

Instead, we should encourage people to check their permissions at the controller/service level. I've always done it at the controller level, and it was super nice. Very easy to see which permissions were needed, very easy to test...

** Removed dependencies **: zend\mvc

## Removing views

As a package that would be API driven does not need the views anymore.

** Removed dependencies ** : zend\view

## Removing collector and various ZfTool

No longer really needed, and we don't have yet a tool for that in middleware's world.

## Authentication

ZfcRbac would no longer rely on Zend\Authentication. Instead, the `isGranted` signature would be changed so that the first parameter is an `IdentityInterface`:

```php
interface AuthorizationServiceInterface
{
    public function isGranted(IdentityInterface $identity, $permission, $context = null);
}
```

Consumer would be responsible to extract it. In PSR-7, each request can be set attribute, for instance here is a possible controller:

```php
public function fooAction(ServerRequestInterface $request)
{
   $identity = $request->getAttribute('logged_user');

   if ($this->authService->isGranted($identity, 'my_perm', ['context' => 'bar']) {

   }
}
```

Maybe we could provide some simple, common interfaces for retrieving logged identity, I'm not so sure.

ping @danizord @weierophinney

danizord:
@bakura10 awesome! I was a bit unsure about the future of Rbac component and ZfcRbac in this middleware age, but your ideas motivated me again :D

Absolutely :+1:, finally we can have `AuthorizationService` stateless.

We could provide some middleware abstraction to replace 2.x guards:

```php
[...]

PR #308, opened 2015-11-05
Start deleting

bakura10:
Hi everyone (special ping to @danizord and @arekkas )

This is the first step of my cleaning work as explained in https://github.com/ZF-Commons/zfc-rbac/issues/307

As said there, the goal is to make zfc-rbac no longer a ZF2 module, but a simpler library based on middleware paradigm (ZfcRbac 2.x will still be supported as a ZF2 library).

As a consequence, a lot of features do not make sense anymore. Guards for now will be removed but we may rethought them around middleware paradigm.

The first step of this PR is to simply remove features and make the tests pass.


bakura10:
Damn, making the tests pass will be a huge mess considering that we wrote them by trying to initialize a ZF2 module (...).

Therefore I think I'll need to merge as it and fix that in a future PR.

danizord:
module.config.php could be removed as well?

bakura10:
You're right!

Issue #240, opened 2014-06-14
[RFC] Assertion refactor


aeneasr:
Hi,

this RFC is related to #181 but I wanted to make a clean approach.
## The problem
- Assertions need additional data to process the logic. This explicitly includes the requested permission.
## Musts
- The AuthorizationService must remain stateless
## A possible solution

``` php
interface ContextAwareAssertionInterface
{
    public function assert(AuthorizationServiceInterface $service, AuthorizationContextInterface $context);
}
```

``` php
interface AuthorizationContextInterface
{
    public function getPermission();
    public function getContext();
}
```

``` php
class AuthorizationContext implements AuthorizationContextInterface
{
     public function __construct($context, $permission)
     {
          // ...
     }

     // ...
}
```

``` php
    protected function assert($assertion, AuthorizationContextInterface $authorizationContext)
    {
        $context = $authorizationContext->getContext();

        if (is_callable($assertion)) {
            return $assertion($this, $context);
        } elseif ($assertion instanceof AssertionInterface) {
            return $assertion->assert($this, $context);
        } elseif ($assertion instanceof ContextAwareAssertionInterface) {
            return $assertion->assert($this, $authorizationContext);
        } elseif (is_string($assertion)) {
            $assertion = $this->assertionPluginManager->get($assertion);

            return $assertion->assert($this, $result);
        }
    }
```

PR #309, opened 2015-11-08
Remove permission

bakura10:
This PR updates dependency to new RBAC (in development).

It also completely removes the PermissionInterface. The only requirement is that the permission given to AuthorizationService is or can be casted to string.

Actually I think you're right @danizord . We could enforce permission as string, as the parameter bag could be added as part of the context, actually.


bakura10:
What do you think of actually also passing the context to the actual "hasPermission" method of the role?

bakura10:
Or maybe embracing the concept explained here: https://github.com/ZF-Commons/zfc-rbac/issues/240

Rbac (not ZfcRbac) would add a new "AuthorizationContextInterface" instead of requiring a string permission?

PR #312, opened 2015-11-08
[WIP] Update factory tests

bakura10:
Some factories proved to be not useful anymore, as we likely remove the dependency to Zend\Authentication and will instead pass the identity as part of the AuthorizationService is granted.


bakura10:
I'm wondering if we should still include the "ObjectRepositoryRoleProvider". I've actually never used it anymore because everything is loaded from the identity automatically. Thoughts @danizord ?

bakura10:
@danizord , while trying to re-understand the code, I can see that the "matchIdentityRoles" is no longer used now. It used to be needed because of ZendDeveloperTools, but now that it has gone, it does not seem to be needed. Any idea of why it could be useful now?

bakura10:
I'm not sure to understand how the flow would work actually. The current identity is provided by the "RoleService", which itself has a dependency to an IdentityProviderInterface (which used to use Zend\Authentication).

Because we are using a middleware approach, I'd think that the good approach would be to have a middleware that would do the authentication logic and store the logged user (if any) as part of the request attributes. However how the RoleService would retrieve the request to get the identity?

The workflow is a bit confusing for me :(.

bakura10:
I'm going to merge that (not passing yet, but I don't want too big PR).

PR #315, opened 2015-11-19
WIP a middleware identity provider for zend authentication service + tests

basz:
proposal


src/Middleware/ZendAuthenticationServiceMiddleware.php
```diff
@@ -0,0 +1,52 @@
...
+     * @param ServerRequestInterface $request
+     * @param ResponseInterface      $response
+     * @param callable|null          $next
+     * @return ResponseInterface
+     */
+    public function __invoke(ServerRequestInterface $request, ResponseInterface $response, callable $next = null)
+    {
+        if ($request->getAttribute($this->attributeName) === null && $this->authService->hasIdentity()) {
+            $identity = $this->authService->getIdentity();
+            $request  = $request->withAttribute($this->attributeName, $identity);
```
bakura10: I think that's where we should set the guest role ?

basz: Nor sure - if we delegate that to the auth service this opens up the idea to stack multiple middleware like this one. If none find an identity null is fed into the isgranted method and guest should be assumed.

bakura10:
Hey @basz ! @Orkin wants to use ZfcRbac in an Expressive app and want to take the lead on the net steps. Maybe you could try to join forces and make some progress on the port to Expressive?

basz:
Ah, now I see this comment. :)

That great news Orkin! Please grab my PR as you see fit. I am sort of planning to use this in an expressive app (a prooph app https://github.com/prooph/service-bus-zfc-rbac-bridge ) too after my holidays so I definitely will be able to do some testing. Unfortunately I need to point my focus on the actual application itself and will be using what works at the moment)

bakura10:
Tests still not passing :)

basz:
Are you taking this PR further @Orkin ?

Orkin:
Yes why not but I don't know where you want to go with :D

basz:
Same boat! I have no idea what the desired direction is for this repo. Removing state is one of the stated goals I believe. But removing assertions isn't part of the plan.

I have no direct involvement and only started this PR to gain some insights in how the middleware idea worked. Then things stalled :)
But I presume you must have some idea as you created a PR specifically against my PR?

Additionally I have no time to continue this ATM.

Bas

basz:
Hmm now i need to make some time as i need it in an expressive app too. :) I review Monday and do a proposal. I think most work is done already.

basz:
IMO this PR should be closed in favour of 316 which I now have working in an application. If we want to provide convenience tooling such a Middleware we're create a new PR from the most up to date dev state.

PR #320, opened 2016-01-04
Added ability to set multiple assertions and their condition for permissions

DavidHavl:
An update to V2 to allow set multiple assertions as well as their condition for permissions.

### DIFF
--- a/src/ZfcRbac/Service/AuthorizationService.php
+++ b/src/ZfcRbac/Service/AuthorizationService.php
@@ -71,8 +71,8 @@ class AuthorizationService implements AuthorizationServiceInterface
     /**
      * Set an assertion
      *
-     * @param string|PermissionInterface         $permission
-     * @param string|callable|AssertionInterface $assertion
+     * @param string|PermissionInterface               $permission
+     * @param string|callable|AssertionInterface|array $assertion
      * @return void
      */
     public function setAssertion($permission, $assertion)
@@ -131,11 +131,56 @@ class AuthorizationService implements AuthorizationServiceInterface
             return false;
         }
 
-        if ($this->hasAssertion($permission)) {
-            return $this->assert($this->assertions[(string) $permission], $context);
+        if (!$this->hasAssertion($permission)) {
+            return true;
         }
 
-        return true;
+        // multiple assertions
+        if (is_array($this->assertions[(string)$permission])) {
+
+            $map = $this->assertions[(string)$permission];
+
+            if (empty($map['assertions'])) {
+                return true;
+            }
+
+            if (!is_array($map['assertions'])) {
+                // convert single assertion to array
+                $map['assertions'] = [$map['assertions']];
+            }
+
+            $condition = isset($map['condition'])
+                ? $map['condition']
+                : AssertionInterface::CONDITION_AND;
+
+            if (AssertionInterface::CONDITION_AND === $condition) {
+                foreach ($map['assertions'] as $assertion) {
+                    if (!$this->assert($assertion, $context)) {
+                        return false;
+                    }
+                }
+
+                return true;
+            }
+
+            if (AssertionInterface::CONDITION_OR === $condition) {
+                foreach ($map['assertions'] as $assertion) {
+                    if ($this->assert($assertion, $context)) {
+                        return true;
+                    }
+                }
+
+                return false;
+            }
+
+            throw new Exception\InvalidArgumentException(sprintf(
+                'Condition must be either "AND" or "OR", %s given',
+                is_object($condition) ? get_class($condition) : gettype($condition)
+            ));
+        } else { // single assertion
+            return $this->assert($this->assertions[(string)$permission], $context);
+        }
+
     }
 
     /**
--- a/docs/06. Using the Authorization Service.md
+++ b/docs/06. Using the Authorization Service.md
@@ -187,6 +187,30 @@ return [
 
 Now, everytime you check for `myPermission`, `myAssertion` will be checked as well.
 
+The assertion map accepts either a single or multiple assertions.
+
+```php
+return [
+    'zfc_rbac' => [
+        'assertion_map' => [
+            'myPermission'  => 'myAssertion', // single assertion
+            'myPermission2' => [              // multiple assertions
+                'assertions' => [
+                    'myAssertion'
+                    'myAssertion2'
+                ],
+                'condition' => \ZfcRbac\Assertion\AssertionInterface::CONDITION_AND
+            ]
+        ]
+    ]
+];
+```
+
+For multiple assertions you can also specify condition AND or OR.
+If 'AND' condition is specified (this is default) all of the assertions must pass the check.
+If 'OR' condition is specified at least one of the assertions must pass the check.
+
+
 ### Checking permissions in a service
 
 So let's check for a permission, shall we?
--- a/tests/ZfcRbacTest/Asset/SimpleFalseAssertion.php
+++ b/tests/ZfcRbacTest/Asset/SimpleFalseAssertion.php
@@ -16,35 +16,33 @@
  * and is licensed under the MIT license.
  */
 
-namespace ZfcRbac\Assertion;
+namespace ZfcRbacTest\Asset;
 
+use ZfcRbac\Assertion\AssertionInterface;
 use ZfcRbac\Service\AuthorizationService;
 
-/**
- * Interface that you can implement for dynamic assertions
- *
- * @author  Michaël Gallego <mic.gallego@gmail.com>
- * @author  Aeneas Rekkas
- * @author  Daniel Gimenes  <daniel@danielgimenes.com.br>
- * @licence MIT
- */
-interface AssertionInterface
+class SimpleFalseAssertion implements AssertionInterface
 {
-
     /**
-     * Condition constants
+     * @var bool
      */
-    const CONDITION_OR  = 'OR';
-    const CONDITION_AND = 'AND';
+    protected $called = false;
 
     /**
-     * Check if this assertion is true
-     *
-     * @TODO: for v3, update the interface to typehint to AuthorizationServiceInterface instead
-     *
-     * @param  AuthorizationService $authorizationService
-     * @param  mixed                $context
+     * {@inheritDoc}
+     */
+    public function assert(AuthorizationService $authorization, $context = null)
+    {
+        $this->called = true;
+
+        return false;
+    }
+
+    /**
      * @return bool
      */
-    public function assert(AuthorizationService $authorizationService);
+    public function getCalled()
+    {
+        return $this->called;
+    }
 }
--- a/tests/ZfcRbacTest/Asset/SimpleTrueAssertion.php
+++ b/tests/ZfcRbacTest/Asset/SimpleTrueAssertion.php
@@ -16,35 +16,33 @@
  * and is licensed under the MIT license.
  */
 
-namespace ZfcRbac\Assertion;
+namespace ZfcRbacTest\Asset;
 
+use ZfcRbac\Assertion\AssertionInterface;
 use ZfcRbac\Service\AuthorizationService;
 
-/**
- * Interface that you can implement for dynamic assertions
- *
- * @author  Michaël Gallego <mic.gallego@gmail.com>
- * @author  Aeneas Rekkas
- * @author  Daniel Gimenes  <daniel@danielgimenes.com.br>
- * @licence MIT
- */
-interface AssertionInterface
+class SimpleTrueAssertion implements AssertionInterface
 {
-
     /**
-     * Condition constants
+     * @var bool
      */
-    const CONDITION_OR  = 'OR';
-    const CONDITION_AND = 'AND';
+    protected $called = false;
 
     /**
-     * Check if this assertion is true
-     *
-     * @TODO: for v3, update the interface to typehint to AuthorizationServiceInterface instead
-     *
-     * @param  AuthorizationService $authorizationService
-     * @param  mixed                $context
+     * {@inheritDoc}
+     */
+    public function assert(AuthorizationService $authorization, $context = null)
+    {
+        $this->called = true;
+
+        return true;
+    }
+
+    /**
      * @return bool
      */
-    public function assert(AuthorizationService $authorizationService);
+    public function getCalled()
+    {
+        return $this->called;
+    }
 }
--- a/tests/ZfcRbacTest/Service/AuthorizationServiceTest.php
+++ b/tests/ZfcRbacTest/Service/AuthorizationServiceTest.php
@@ -78,17 +78,120 @@ class AuthorizationServiceTest extends \PHPUnit_Framework_TestCase
                 ]
             ],
 
-            // Simple is accepted from assertion map
+            // Simple is accepted with empty assertions
             [
                 'admin',
                 'delete',
                 true,
                 true,
                 [
-                    'delete' => 'ZfcRbacTest\Asset\SimpleAssertion'
+                    'delete' => []
+                ]
+            ],
+            // Simple is accepted from assertion map multiple
+            [
+                'admin',
+                'delete',
+                true,
+                true,
+                [
+                    'delete' => [
+                        'assertions' => 'ZfcRbacTest\Asset\SimpleAssertion'
+                    ]
+                ]
+            ],
+            // Simple is accepted from assertion map multiple
+            [
+                'admin',
+                'delete',
+                true,
+                true,
+                [
+                    'delete' => [
+                        'assertions' => 'ZfcRbacTest\Asset\SimpleAssertion'
+                    ]
                 ]
             ],
 
+            // Simple is refused from assertion map with multiple assertions default condition
+            [
+                'admin',
+                'delete',
+                null,
+                false,
+                [
+                    'delete' => [
+                        'assertions' => [
+                            'ZfcRbacTest\Asset\SimpleTrueAssertion',
+                            'ZfcRbacTest\Asset\SimpleFalseAssertion'
+                        ]
+                    ]
+                ]
+            ],
+            // Simple is refused from assertion map with multiple assertions, condition AND
+            [
+                'admin',
+                'delete',
+                null,
+                false,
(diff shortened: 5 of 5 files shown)


bakura10:
I like the idea. What about bringing a new "AssertionSet" class that would contain the two constants, and abstract the logic of checking all the assertions and returning a boolean, so the `isGranted` can stay simple?

DavidHavl:
That actually sounds good. I'll work on it within next few days.

lorenzoferrarajr:
This is a very good idea! It would help us keep things better organized.

basz:
@DavidHavl I like this idea! Now work on V3 is progressing it would be bring this feature there too.

DavidHavl:
Thanks @basz, I have been extremely busy with work past few months but I should have more time now for this.

prolic:
Is there still progress on this?

src/ZfcRbac/Service/AuthorizationService.php
```diff
@@ -122,19 +122,15 @@ public function getIdentity()
     public function isGranted($permission, $context = null)
     {
         $roles = $this->roleService->getIdentityRoles();
-
```
basz: no need to remove these empty lines

prolic:
Also coverage decreased, can you add some more tests, please?

DavidHavl:
@basz that was my initial idea but it was not accepted. 
I'll work on it.

basz:
Ok, remember why? Perhaps was fora good reason...

src/ZfcRbac/Service/AuthorizationService.php
```diff
@@ -150,10 +177,6 @@ protected function assert($assertion, $context = null)
             return $assertion($this, $context);
         } elseif ($assertion instanceof AssertionInterface) {
             return $assertion->assert($this, $context);
-        } elseif (is_string($assertion)) {
```
svycka: why remove this?

src/ZfcRbac/Service/AuthorizationService.php
```diff
@@ -72,11 +73,34 @@ public function __construct(Rbac $rbac, RoleService $roleService, AssertionPlugi
...
      *
      * @param string|PermissionInterface         $permission
-     * @param string|callable|AssertionInterface $assertion
+     * @param string|callable|array|AssertionInterface $assertion
      * @return void
      */
     public function setAssertion($permission, $assertion)
     {
+        // if is name of the assertion, retrieve an actual instance from assertion plugin manager
+        if (is_string($assertion)) {
```
svycka: it's slower to do it here instead of where it was in `assert()` method because it will be created on every request even if this is not used.

src/ZfcRbac/Service/AuthorizationService.php
```diff
@@ -72,11 +73,34 @@ public function __construct(Rbac $rbac, RoleService $roleService, AssertionPlugi
...
+            $assertion = $this->assertionPluginManager->get($assertion);
+        } elseif (is_array($assertion)) { // else if multiple assertion definition, create assertion set.
```
svycka: do we need this whole block here? can't we move it to AssertionSet? something like:
```php
if (is_array($assertion)) {
   $assertion = new AssertionSet($assertion, $this->assertionPluginManager);
}
```

src/ZfcRbac/Service/AuthorizationService.php
```diff
@@ -72,11 +73,34 @@ public function __construct(Rbac $rbac, RoleService $roleService, AssertionPlugi
...
+            if (!isset($assertion['assertions'])) {
+                $assertion['assertions'] = (array)$assertion;
+            } elseif (!is_array($assertion['assertions'])) {
+                $assertion['assertions'] = (array)$assertion['assertions'];
+            }
+
+            // retrieve an actual instance from assertion plugin manager if necessary
+            foreach ($assertion['assertions'] as $key => $value) {
+                if (is_string($value)) {
+                    $assertion['assertions'][$key] = $this->assertionPluginManager->get($value);
```
svycka: fetching assertions would be better when they are really needed in `assert()` method in this case `AssertionSet::assert()`

src/ZfcRbac/Service/AuthorizationService.php
```diff
@@ -88,7 +112,10 @@ public function setAssertion($permission, $assertion)
      */
     public function setAssertions(array $assertions)
     {
-        $this->assertions = $assertions;
```
svycka: don't know if this is good creating assertions while setting them. Maybe we could create AssertionSet only if we need in `assert()` method?

src/ZfcRbac/Service/AuthorizationService.php
```diff
@@ -150,10 +177,6 @@ protected function assert($assertion, $context = null)
             return $assertion($this, $context);
         } elseif ($assertion instanceof AssertionInterface) {
             return $assertion->assert($this, $context);
-        } elseif (is_string($assertion)) {
-            $assertion = $this->assertionPluginManager->get($assertion);
-
-            return $assertion->assert($this, $context);
         }
```
svycka: maybe add this:
```php
elseif (is_array($assertion)) {
    $assertionSet = new AssertionSet($assertion, $this->assertionPluginManager);
    return $assertionSet->assert($this, $context);
}
```
see my above comments why

src/ZfcRbac/Assertion/AssertionSet.php
```diff
@@ -0,0 +1,200 @@
...
+    /**
+     * Check if assertions are successful
+     *
+     * @param  AuthorizationService $authorizationService
+     * @param  mixed                $context
+     * @return bool
+     */
+    public function assert(AuthorizationService $authorizationService, $context = null)
+    {
+        if (self::CONDITION_AND === $this->condition) {
```
svycka: if this class not a final then maybe `AssertionSet::CONDITION_AND`

src/ZfcRbac/Assertion/AssertionSet.php
```diff
@@ -0,0 +1,200 @@
...
+            foreach ($this->assertions as $assertion) {
+                if (!$assertion->assert($authorizationService, $context)) {
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
+        if (self::CONDITION_OR === $this->condition) {
```
svycka: if this class not a final then maybe `AssertionSet::CONDITION_OR`

src/ZfcRbac/Service/AuthorizationService.php
svycka: one more thought. Maybe when moved to `AssertionSet` we could have nested assertions something like this:
```
return [
    'zfc_rbac' => [
        'assertion_map' => [
             // single assertion
            'myPermission'  => 'myAssertion',
            'myPermission2' => [ 
                // multiple assertions
                'assertions' => [
                    'myAssertion',
                    'myAssertion2',
                    [ 
                      // multiple assertions
                      'assertions' => [
                          'myAssertion3',
                          'myAssertion4'
                      ], 
                      // condition
                      'condition' => \ZfcRbac\Assertion\AssertionSet::CONDITION_OR
                  ]
                ], 
                // condition
                'condition' => \ZfcRbac\Assertion\AssertionSet::CONDITION_AND
            ]
        ]
    ]
];
```

src/ZfcRbac/Assertion/AssertionSet.php
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    /**
+     * Check if assertions are successful
+     *
+     * @param  AuthorizationService $authorizationService
+     * @param  mixed                $context
+     * @return bool
+     */
+    public function assert(AuthorizationService $authorizationService, $context = null)
```
DavidHavl: Good point. Thanx.

src/ZfcRbac/Assertion/AssertionSet.php
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    /**
+     * Returns true if the assertion if defined.
+     *
+     * @param string $name The assertion name
+     *
+     * @return bool true if the assertion is defined, false otherwise
+     */
+    public function hasAssertion($name)
```
DavidHavl: Yes people can when they use the class directly (see docs).

src/ZfcRbac/Assertion/AssertionSet.php
```diff
@@ -0,0 +1,200 @@
...
+use ZfcRbac\Service\AuthorizationService;
+
+/**
+ * Assertion set to hold and process multiple assertions
+ *
+ * @author  David Havl
+ * @licence MIT
+ */
+
+class AssertionSet implements AssertionInterface, \IteratorAggregate
```
svycka: I don't think anyone would want to modify this after it is added to `AuthorizationService` and even if they want this is not a good practice.
@basz @prolic @bakura10 what you think?

src/ZfcRbac/Assertion/AssertionSet.php
svycka: okey maybe they could but still not convinced. Even if they want to modify it what is bad I think they could create new one instead. because reusing same set on different permissions and randomly modify them at the same time can lead to big problems. And if users would want those evil methods they could extend the class anyway no? so why add those we don't want to maintain +6 methods for just in case users would need them.
but again @basz @prolic @bakura10 @danizord what you think?

src/ZfcRbac/Service/AuthorizationService.php
svycka: and I don't like idea adding `AssertionSet` logic inside `AuthorizationService` while it is just simple custom assertion.
but again @bakura10 @prolic @basz @danizord

src/ZfcRbac/Assertion/AssertionSet.php
DavidHavl: Not sure I understand. I was imagining usage of it for example like this: 

Creating a factory where one can do anything they want
```
use Interop\Container\ContainerInterface;
use Zend\ServiceManager\FactoryInterface;
use Zend\ServiceManager\ServiceLocatorInterface;
use ZfcRbac\Assertion\AssertionSet;

class MyAssertionSetFactory implements FactoryInterface
{
    /**
     * {@inheritDoc}
     *
     * @return AssertionSet
     */
    public function __invoke(ContainerInterface $container, $name, array $options = null)
    {
        $assertionManager = $container->get('ZfcRbac\Assertion\AssertionPluginManager');
        $assertion1 = $assertionManager->get('myAssertion1');
        $assertion2 = $assertionManager->get('myAssertion2');

        // create instance, set condition and add assertions
        $assertionSet = new AssertionSet();
        $assertionSet->setCondition(AssertionSet::CONDITION_OR);
        $assertionSet->setAssertions([$assertion1, $assertion2]);
        return $assertionSet;
    }
    
    /**
     * {@inheritDoc}
     *
     * For use with zend-servicemanager v2; proxies to __invoke().
     *
     * @param ServiceLocatorInterface $container
     * @return \ZfcRbac\Assertion\AssertionSet
     */
    public function createService(ServiceLocatorInterface $container)
    {
        // Retrieve the parent container when under zend-servicemanager v2
        if (method_exists($container, 'getServiceLocator')) {
            $container = $container->getServiceLocator() ?: $container;
        }

        return $this($container, AssertionSet::class);
    }
}

```



And then add it to assertion manager and assertion map config:

```php
return [
    'zfc_rbac' => [
        'assertion_manager' => [
            'factories' => [
                'myAssertionSet' => MyAssertionSetFactory::class
            ]
        ],

        'assertion_map' => [
            'myPermission'  => 'myAssertion', // single assertion
            'myPermission2' => 'myAssertionSet' // multiple assertions in set
        ]
    ]
];
```

svycka: you could do the same without those methods:
```php
$assertionManager = $container->get('ZfcRbac\Assertion\AssertionPluginManager');
$assertion1 = $assertionManager->get('myAssertion1');
$assertion2 = $assertionManager->get('myAssertion2');

// create instance, set condition and add assertions
 $assertionSet = new AssertionSet([
    'assertions' => [$assertion1, $assertion2],
    'condition' => AssertionSet::CONDITION_OR
]);
```
and this would prevent from changing later.

src/ZfcRbac/Service/AuthorizationService.php
DavidHavl: I understand your concern, but look at the contradicting response/suggestion from bakura10 few months ago... this was the initial reasoning behind adding it and I am very fond of it since then. In fact I am using the concept now on several projects (on tweaked ZfcRbac module) and I like the versatility of that solution. But I am opened to suggestions.

src/ZfcRbac/Assertion/AssertionSet.php
danizord: +1 for immutable approach. That makes unit testing way easier and confident since you don't need to test a lot of possible different states.

DavidHavl: Fair enough. That is a good point. 
Will rewrite.
Thanks.

src/ZfcRbac/Service/AuthorizationService.php
svycka: sorry but I don't understand I just say this block is almost same as `AssertionSet::setAssertions()` so why duplicate it here. Also I understand why you don't want AssertionPluginManager in AssertionSet but thats also not so logical because of this you can't use assertion names while creating AssertionSet you have to first fetch all required assertions then add to set this is performance hungry if you have a lot of assertions.

danizord: What about the following approach?
- `assertion_map` config key accepts only `string => string` (permission name => assertion name)
- Another config key for `assertion_sets` that accepts `string => array` (assertion name => assertion config)
- A default `AssertionSetFactory` implementation that reads the `assertion_sets` config key and builds `AssertionSet`
- `AuthorizationService` always fetch assertions from `AssertionPluginManager`

src/ZfcRbac/Assertion/AssertionSet.php
svycka: I said but will repeat:  not good that we can't do this:
```php
$assertionSet = new AssertionSet([
    'assertions' => ['myAssertion1', 'myAssertion2'],
    'condition' => AssertionSet::CONDITION_OR
]);
```

DavidHavl: @svycka great, I am working on it.

DavidHavl:
@danizord the approach you describe is interesting, but I have a concern from usability point of view. I mean, I think it may be more confusing and less intuitive for people wanting to use the module to have yet another separate config for sets. Plus I am not sure about performance implications either.

src/ZfcRbac/Assertion/AssertionSet.php
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /**
+     * Check if assertions are successful
+     *
+     * @param  AuthorizationService $authorizationService
+     * @param  mixed                $context
+     * @return bool
+     */
+    public function assert(AuthorizationService $authorizationService, $context = null)
```
DavidHavl: Done

src/ZfcRbac/Service/AuthorizationService.php
DavidHavl: This is now back, you are right, it makes it more performant to have the retrieval only when it's called/needed.

svycka:
First yes you moved `AssertionSet` creation to `assert()` but this still not optimal you still creating all assertions even if they are not required. Imagine you have 20 assertions with `or` condition so you have to create all 20 when in reality you should have created just first one and all others are not required.

I don't like that `AssertionSet` does not support assertions as `string` or `callback` while supported are `string|callable|array|AssertionInterface` and only allows `AssertionInterface` at the same time does not check if `$assertion` is `AssertionInterface`. Also do we really need names for assertions in config?

And I still would like to change `AuthorizatrionService::assert()` to something like:
```php
protected function assert($assertion, $context = null)
{
    $assertion = new AssertionSet($assertion, $this->assertionPluginManager);
    return $assertion->assert($this, $context);
}
```
this is also `AssertionSet` and if I understand correctly should work no? haven't tested
```php
return [
    'zfc_rbac' => [
        'assertion_map' => [
             // single assertion
            'myPermission'  => 'myAssertion',
            // assertion set with default condition `and`
            'myPermission2' => [ 
                'myAssertion',
                'myAssertion2',
            ]
        ]
    ]
];
```

or if you don't like adding `AssertionPluginManager` to `AssertionSet` then maybe introduce new method in AuthorizationService:
```php
public function getAssertion(string $assertion)
{
    return $this->assertionPluginManager->get($assertion);
}
```

DavidHavl:
@svycka I see your point in case of `OR` condition where it is redundant to create rest of the assertions.
However, the simplified `AuthorizatrionService::assert()` method you propose has a flow. It would have to create an instance of AssertionSet every time even for single standalone assertions or callables or named assertion which is redundant.

svycka:
It is not just or condition with and this also valid if first fail all
others does not mater and will not be created. Also if you have concerns
about creating this object then ok you can create it if assertion is array.
But if everyone agrees with you then ok let's leave it this way.

DavidHavl:
Hey guys, I don't think I will have enough time to do more changes on this in next few months (got an urgent project that doesn't use this), so feel free to do a pull 
and then if you think it needs more refinements feel free to adjust it.

src/ZfcRbac/Assertion/AssertionSet.php
```diff
@@ -0,0 +1,113 @@
...
+use ZfcRbac\Exception\InvalidArgumentException;
+use ZfcRbac\Service\AuthorizationService;
+
+/**
+ * Assertion set to hold and process multiple assertions
+ *
+ * @author  David Havl
+ * @licence MIT
+ */
+class AssertionSet implements AssertionInterface
```
prolic: let's mark final

basz:
the author expressed he didn't have time to continue it. We now have this functionality in the develop branch and #379 could be backported to master if anyone needs it. (hence the label change)

DavidHavl:
Great, thanks for finishing it up @basz !

PR #321, opened 2016-02-09
Fixed conditions in ControllerPermissionsGuard

Itach1Uchixa:
Old repository hasn't condition check in ControllerPermissionsGuard


bakura10:
Hi,

The tests are failing. This is actually quite a small BC: not specifying condition type before would throw exception, now it's assuming "AND". I'd say that I'd prefer to be explicit on something as importnat as security, if you want "AND", you should specify it in your config, rather than relying on the module to do a default choice for you.

Itach1Uchixa:
Hi, i understand but i want 'ControllerPermissionsGuard' to be like 'RoutePermissionsGuard', because 'RoutePermissionsGuard' takes and condition as default as written in docs

bakura10:
That makes sense, I'll likely merge that as a 2.6.0, the BC seems very very minor. However you'll need to fix the tests before I can merge this PR.

prolic:
@basz Can you take a look at this? We also need to decide whether this is good for 2.6 or if a 2.7 is required.

src/ZfcRbac/Guard/ControllerPermissionsGuard.php
```diff
@@ -130,12 +134,35 @@ public function isGranted(MvcEvent $event)
             return true;
         }
 
-        foreach ($allowedPermissions as $permission) {
-            if (!$this->authorizationService->isGranted($permission)) {
-                return false;
+        // In old repository there wasn't condition check
+        // New lines of code is below
```
svycka: these comments does not feel usefull

src/ZfcRbac/Guard/ControllerPermissionsGuard.php
```diff
@@ -130,12 +134,35 @@ public function isGranted(MvcEvent $event)
...
+        $condition = $this->rules[$controller][1];
```
svycka: maybe add comment explaining what `[1]` means.

basz:
on a quick install and review, this looks ok to me...

2.7 I would say...

svycka:
@prolic if it's not a fix then 2.7

Itach1Uchixa:
@svycka I've added comments that you required and also cleaned commits mess

prolic:
accidentally closed or on purpose?

Issue #322, opened 2016-03-15
InMemoryRoleProvider can reuse role objects


Wilt:
In the current setup the child role in a `HierarchicalRole` is another Role instance with the same name. This is because of how the `InMemoryRoleProvider` is creating the roles. For example [the `$memberRole` in the `InMemoryRoleProviderTest`](https://github.com/ZF-Commons/zfc-rbac/blob/master/tests/ZfcRbacTest/Role/InMemoryRoleProviderTest.php#L53) is not the same object as the first child role of the `$adminRole`.

It would be cool to change the setup so that these tests are successful. I also think it will save some memory because in this case the roles are instantiated only once.

    assertTrue($adminRole->getChildren()['member'] === $memberRole);
    assertTrue($memberRole->getChildren()['guest'] === $guestRole);

danizord:
Looks good for me, can you provide a PR @Wilt?

Wilt:
Here a [PR](https://github.com/ZF-Commons/zfc-rbac/pull/323)

danizord:
Fixed by #325

PR #323, opened 2016-03-16
Improvement of InMemoryRoleProvider 

Wilt:
This PR follows on the issue [here](https://github.com/ZF-Commons/zfc-rbac/issues/322)

A `$roles` array storage is added to the InMemoryRoleProvider in which resolved roles are stored. It means role instances with the same name are reused (and not recreated). This will save memory and allows 1 to 1 role object comparison since the `InMemoryRoleProvider` will always return the same object instance for a certain `$roleName`.

A test for this enhancement is provided.


Wilt:
I will close and change target to develop branch

PR #324, opened 2016-03-16
Improvement of InMemoryRoleProvider

Wilt:
This PR follows on the issue [here](https://github.com/ZF-Commons/zfc-rbac/issues/322)

A `$roles` array storage is added to the InMemoryRoleProvider in which resolved roles are stored. It means role instances with the same name are reused (and not recreated). This will save memory and allows 1 to 1 role object comparison since the `InMemoryRoleProvider` will always return the same object instance for a certain `$roleName`.

A test for this enhancement is provided.


Wilt:
Failing tests are not because of my changes...

Wilt:
Failing tests are not because of my changes...

src/Role/InMemoryRoleProvider.php
```diff
@@ -62,19 +72,35 @@ public function getRoles(array $roleNames)
...
+     */
+    protected function getRole($roleName)
+    {
+        if (isset($this->roles[$roleName])) {
+            return $this->roles[$roleName];
+        }
+
+        // If no config, we create a simple role with no permission
+        if (!isset($this->rolesConfig[$roleName])) {
+            $role = new Role($roleName);
```
danizord: Return early and avoid the else keyword

test/Role/InMemoryRoleProviderTest.php
```diff
@@ -64,5 +64,8 @@ public function testInMemoryProvider()
         $this->assertEquals('guest', $guestRole->getName());
         $this->assertFalse($guestRole->hasPermission('write'));
         $this->assertFalse($guestRole->hasPermission('delete'));
+
+        $this->assertTrue($adminRole->getChildren()['member'] === $memberRole);
+        $this->assertTrue($memberRole->getChildren()['guest'] === $guestRole);
```
danizord: Can use assertSame instead

danizord:
Looks good for me, :+1:

@bakura10?

Wilt:
@danizord Updated code following your comments

bakura10:
Travis tests are not passing anymore. Actually you have made this PR against the develop branch which is a new major version that relies on a new version of Rbac.

I think that what you want is doing the PR against master.

Wilt:
@bakura10 Sure, I never seem to do it right. Others want `develop` as target branch :S
I will change it...

Wilt:
Now here: https://github.com/ZF-Commons/zfc-rbac/pull/325

PR #327, opened 2016-03-22
changes "@licence" with "@license"

manuakasam:
The proper docblock is "@license" and it needs to be proper so that annotation parsing libraries are skipping this annotation as it is a phpdoc one.

Just ran into an issue where we were adding the exception to the logging. Our logger is configured to basically do `if is_object($attachment) : JMSSerialize($attachment)` and as JMS Serializer uses annotations internally to exclude some stuff, we ran into the funny exception

```
[Semantical Error] The annotation "@licence" in class ZfcRbac\Exception\UnauthorizedException was never imported. Did you maybe forget to add a "use" statement for this annotation?`
```

This PR fixes that :)

Kudos to @grizzm0 for noticing the mistake in the first place


bakura10:
?? It was wrong? :p

PR #328, opened 2016-03-22
Changes "@licence" with "@license"

manuakasam:
As pointed out by @grizzm0 the proper docblock is `@license`

We ran into the issue that our logger uses JMS Serializer (which internally uses an annotation parser) to log any kind of objects that get serialized into JSON. However due to the misspelling of licence it tried to find a parser for `@licence` which doesn't exist :)

### DIFF
--- a/src/ZfcRbac/Guard/GuardPluginManager.php
+++ b/src/ZfcRbac/Guard/GuardPluginManager.php
@@ -27,8 +27,12 @@ use ZfcRbac\Exception;
  * @method GuardInterface get($name)
  *
  * @author  Michaël Gallego <mic.gallego@gmail.com>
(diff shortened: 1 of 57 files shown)


bakura10:
Good catch!

bakura10:
I've tagged as 2.5.5. Thanks!

src/ZfcRbac/Guard/GuardPluginManager.php
```diff
@@ -27,8 +27,12 @@
  * @method GuardInterface get($name)
  *
  * @author  Michaël Gallego <mic.gallego@gmail.com>
+<<<<<<< b459677bbe1c328d819969eed3fe7df68cb1112b
```
grizzm0: Accidental conflict message merged. @bakura10

PR #335, opened 2016-06-23
allow zf3

prolic:
* Allow Zend-MVC 3.0 in composer
* Allow Zend-Servicemanager 3.0 in composer
* Allow Zend\Eventmanager 3.0 in composer
* Update to ZendDeveloperTools 1.1 in composer
* Update to DoctrineModule 1.1 in composer
* Update to DoctrineORMModule 0.11 in composer
* drop support of PHP 5.4 & 5.5
* allow PHP 7

some further notes:

1) DoctrineORMModule does not support Zend-MVC 3.0, but this module still can support it, as the tests are showing. If you want to run this module with DoctrineORMModule, then you have to use Zend-MVC 2.7 for now.

2) There are some BC breaks in the factories. They are working the with Zend-Servicemanager v2 and v3 but do not implement the MutableCreationOptionsInterface, f.e. Also the plugin managers get the main service locator injected into the constructor, instead of calling $pluginmanager->setServiceLocator($sm);
This has to be noted in the changelog. You don't need to care about this, when you simply use this module as is. If you extended or aggregated the provided factories, you need to update your code accordingly.


danizord:
Mhmm seems like it is time to drop PHP 5.4 support?

prolic:
Even time to drop 5.5 I would say :-)

danizord:
ok, I'm just affraid that our build does not run tests with both ZF2 and ZF3. I think we could drop ZF2 as well then.

prolic:
Oh, no problem, let me do this... wait...

prolic:
so, that should to the trick.

prolic:
damn. some tests are failing with minimum requirements already. I'll try to fix that tomorrow.

prolic:
Any ideas why latest zend-servicemanager installed is 2.7.6 instead of 3.1? Is there any other dependency requiring the older version perhaps?

prolic:
Damn, currently DoctrineModule and zend-developer-tools do not allow Zend-Servicemanager v3

for DoctrineModule, there is a PR here: https://github.com/doctrine/DoctrineModule/pull/558
for ZendDeveloperTools there is only an issue here: https://github.com/zendframework/ZendDeveloperTools/issues/207

basz:
So you can prepare against that PR... I did that when I wanted to zf3 proof SlmQueue. Whenever doctrine orm is updated I'm fairly sure it will work...

prolic:
Please review carefully. I needed to change a lot of code.

prolic:
ping @danizord

tests/ZfcRbacTest/Util/ServiceManagerFactory.php
```diff
@@ -59,12 +59,11 @@ public static function getApplicationConfig()
...
     {
         $config = $config ?: static::getApplicationConfig();
-        $serviceManager = new ServiceManager(
-            new ServiceManagerConfig(
-                isset($config['service_manager']) ? $config['service_manager'] : []
-            )
-        );
+        $serviceManager = new ServiceManager();
         $serviceManager->setService('ApplicationConfig', $config);
+        $serviceManagerConfig = new ServiceManagerConfig();
```
danizord: Should pass `$config['service_manager']` here?

danizord:
@prolic IIRC SM 2.x already supports callable factories, so why keeping these createService?

tests/ZfcRbacTest/Util/ServiceManagerFactory.php
prolic: done.

prolic:
IIRC SM 2.x already supports callable factories, so why keeping these createService?

We need to keep this, in order to be compatible with ZF2 and ZF3 ServiceManager. The __invoke-method has different arguments in both:

ZF3: $container, $requestedName, $options
ZF2: $serviceLocator, $canonicalName, $requestedName

danizord:
@prolic but for most factories you can drop the interface and use a simple callable with single `ContainerInterface` argument.

prolic:
@danizord No you can't because the library passes options to the container.

Some examples:

1)
$roleProvider = $pluginManager->get('ZfcRbac\Role\ObjectRepositoryRoleProvider', $options);
(tests/ZfcRbacTest/Factory/ObjectRepositoryRoleProviderFactoryTest.php:66)

2)
$guard = $pluginManager->get($type, $options);
(tests/ZfcRbacTest/Guard/GuardPluginManagerTest.php:88)

We could get rid of it in some factories, but not in all. So why bother and make it inconsistent? It's working with ZF2 and ZF3 now, and changing it cannot be done in a consistent way.

I'd say, let's cleanup the factories when we decide to drop support for ZF2, than the factories can be updated to only support ZF3 and the createService-method can be easily removed.

tests/ZfcRbacTest/Factory/IsGrantedViewHelperFactoryTest.php
```diff
@@ -31,16 +31,15 @@ public function testFactory()
...
+        $pluginManager  = new HelperPluginManager($serviceManager);
 
         $serviceManager->setService(
             'ZfcRbac\Service\AuthorizationService',
             $this->getMock('ZfcRbac\Service\AuthorizationServiceInterface')
         );
 
         $factory   = new IsGrantedViewHelperFactory();
-        $isGranted = $factory->createService($pluginManager);
+        $isGranted = $factory->createService($serviceManager);
```
svycka: BC break?

src/ZfcRbac/Factory/IsGrantedViewHelperFactory.php
```diff
@@ -31,14 +35,25 @@
...
+     * @param string $requestedName
+     * @param array|null $options
      * @return IsGranted
      */
-    public function createService(ServiceLocatorInterface $serviceLocator)
+    public function __invoke(ContainerInterface $container, $requestedName, array $options = null)
     {
         /* @var \ZfcRbac\Service\AuthorizationService $authorizationService */
-        $authorizationService = $serviceLocator->getServiceLocator()->get('ZfcRbac\Service\AuthorizationService');
+        $authorizationService = $container->get('ZfcRbac\Service\AuthorizationService');
```
svycka: BC break? and others pluginServiceManager factories

tests/ZfcRbacTest/Factory/ControllerGuardFactoryTest.php
```diff
@@ -50,13 +50,17 @@ public function testFactory()
...
-        $pluginManager = new GuardPluginManager();
-        $pluginManager->setServiceLocator($serviceManager);
+        $pluginManager = new GuardPluginManager($serviceManager);
 
         $factory         = new ControllerGuardFactory();
-        $controllerGuard = $factory->createService($pluginManager);
 
-        $this->assertInstanceOf('ZfcRbac\Guard\ControllerGuard', $controllerGuard);
-        $this->assertEquals(GuardInterface::POLICY_ALLOW, $controllerGuard->getProtectionPolicy());
+        if (method_exists($serviceManager, 'build')) { // servicemanager v3
```
svycka: why do you check this? 
```php
$guard = $factory->createService($pluginManager);
```
should be for both v2 and v3 versions the same or if you really want to test v3 should be:
```php
$guard = $factory($serviceManager, \ZfcRbac\Guard\ControllerGuard::class);
```
should be fixed in all tests

src/ZfcRbac/Factory/ObjectRepositoryRoleProviderFactory.php
```diff
@@ -40,42 +39,55 @@ class ObjectRepositoryRoleProviderFactory implements FactoryInterface, MutableCr
...
         );
     }
+
+    /**
+     * {@inheritDoc}
+     * @return ObjectRepositoryRoleProvider
+     */
+    public function createService(ServiceLocatorInterface $serviceLocator)
+    {
+        return $this($serviceLocator, ObjectRepositoryRoleProvider::class);
```
svycka: I would suggest keep `__invoke()` clean only what required for SMv3 because this will be called only with SMv3 and do all required changes here. For example:
```php
return $this($serviceLocator->getServiceLocator(), ObjectRepositoryRoleProvider::class, $this->options);
```

src/ZfcRbac/Factory/ObjectRepositoryRoleProviderFactory.php
```diff
@@ -40,42 +39,55 @@ class ObjectRepositoryRoleProviderFactory implements FactoryInterface, MutableCr
...
         $this->options = $options;
     }
 
     /**
-     * {@inheritDoc}
+     * @param ContainerInterface $container
      * @return ObjectRepositoryRoleProvider
      */
-    public function createService(ServiceLocatorInterface $serviceLocator)
+    public function __invoke(ContainerInterface $container, $resolvedName, $options)
```
svycka: `$options` param is against interface should be optional

src/ZfcRbac/Factory/RouteGuardFactory.php
```diff
@@ -39,28 +38,40 @@ class RouteGuardFactory implements FactoryInterface, MutableCreationOptionsInter
...
     /**
-     * {@inheritDoc}
+     * @param ContainerInterface $container
      * @return RouteGuard
      */
-    public function createService(ServiceLocatorInterface $serviceLocator)
+    public function __invoke(ContainerInterface $container)
     {
-        $parentLocator = $serviceLocator->getServiceLocator();
+        if (! method_exists($container, 'build')) { // servicemanager v3
```
svycka: this would be not required if `$container->getServiceLocator()` moved to `createService()`

src/ZfcRbac/Guard/AbstractGuard.php
```diff
@@ -46,9 +46,9 @@
     /**
      * {@inheritDoc}
      */
-    public function attach(EventManagerInterface $events)
+    public function attach(EventManagerInterface $events, $priority = self::EVENT_PRIORITY)
```
svycka: maybe `AbstractGuard::EVENT_PRIORITY` would be better?

src/ZfcRbac/Initializer/AuthorizationServiceInitializer.php
```diff
@@ -32,17 +33,26 @@
...
     /**
-     * @see \Zend\ServiceManager\InitializerInterface::initialize()
+     * @param ContainerInterface $container
+     * @param object $instance
      */
-    public function initialize($instance, ServiceLocatorInterface $serviceLocator)
+    public function __invoke(ContainerInterface $container, $instance)
     {
         if ($instance instanceof AuthorizationServiceAwareInterface) {
-            if ($serviceLocator instanceof AbstractPluginManager) {
```
svycka: this can be moved to `initialize()`

tests/ZfcRbacTest/Util/ServiceManagerFactory.php
```diff
@@ -59,12 +59,13 @@ public static function getApplicationConfig()
...
-                isset($config['service_manager']) ? $config['service_manager'] : []
-            )
-        );
+        $serviceManager = new ServiceManager();
         $serviceManager->setService('ApplicationConfig', $config);
+        $serviceManagerConfig = new ServiceManagerConfig(
+            isset($config['service_manager']) ? $config['service_manager'] : []
+        );
+        $serviceManagerConfig->configureServiceManager($serviceManager);
+        $serviceManager->setAllowOverride(true);
```
svycka: maybe better allow only in place we really need this? Not sure we even need this.

svycka:
@prolic you did a lot of changes in tests even covering BC breaks please revert. Also Zend-MVC 3.0 is not possible with DoctrineORMModule at least BC way. So maybe we can drop zf2 support or we can't have zend-mvc 3.0

src/ZfcRbac/Factory/AssertionPluginManagerFactory.php
```diff
@@ -18,30 +18,42 @@
 
 namespace ZfcRbac\Factory;
 
+use Interop\Container\ContainerInterface;
+use Interop\Container\Exception\ContainerException;
```
svycka: not used

src/ZfcRbac/Factory/AssertionPluginManagerFactory.php
```diff
@@ -18,30 +18,42 @@
...
 use Zend\ServiceManager\Config;
+use Zend\ServiceManager\Exception\ServiceNotCreatedException;
```
svycka: not used

src/ZfcRbac/Factory/AssertionPluginManagerFactory.php
```diff
@@ -18,30 +18,42 @@
...
+use Zend\ServiceManager\Exception\ServiceNotFoundException;
```
svycka: not used

src/ZfcRbac/Factory/GuardsFactory.php
```diff
@@ -18,6 +18,10 @@
 
 namespace ZfcRbac\Factory;
 
+use Interop\Container\ContainerInterface;
+use Interop\Container\Exception\ContainerException;
```
svycka: not used

src/ZfcRbac/Factory/GuardsFactory.php
```diff
@@ -18,6 +18,10 @@
...
+use Zend\ServiceManager\Exception\ServiceNotCreatedException;
```
svycka: not used

src/ZfcRbac/Factory/GuardsFactory.php
```diff
@@ -18,6 +18,10 @@
...
+use Zend\ServiceManager\Exception\ServiceNotFoundException;
```
svycka: not used

src/ZfcRbac/Factory/HasRoleViewHelperFactory.php
```diff
@@ -18,6 +18,10 @@
 
 namespace ZfcRbac\Factory;
 
+use Interop\Container\ContainerInterface;
+use Interop\Container\Exception\ContainerException;
```
svycka: not used

src/ZfcRbac/Factory/HasRoleViewHelperFactory.php
```diff
@@ -18,6 +18,10 @@
...
+use Zend\ServiceManager\Exception\ServiceNotCreatedException;
```
svycka: not used

src/ZfcRbac/Factory/HasRoleViewHelperFactory.php
```diff
@@ -18,6 +18,10 @@
...
+use Zend\ServiceManager\Exception\ServiceNotFoundException;
```
svycka: not used

src/ZfcRbac/Factory/IsGrantedPluginFactory.php
```diff
@@ -18,6 +18,10 @@
 
 namespace ZfcRbac\Factory;
 
+use Interop\Container\ContainerInterface;
+use Interop\Container\Exception\ContainerException;
```
svycka: not used

src/ZfcRbac/Factory/IsGrantedPluginFactory.php
```diff
@@ -18,6 +18,10 @@
...
+use Zend\ServiceManager\Exception\ServiceNotCreatedException;
```
svycka: not used

src/ZfcRbac/Factory/IsGrantedPluginFactory.php
```diff
@@ -18,6 +18,10 @@
...
+use Zend\ServiceManager\Exception\ServiceNotFoundException;
```
svycka: not used

src/ZfcRbac/Factory/IsGrantedPluginFactory.php
```diff
@@ -31,14 +35,25 @@
...
+     * @param ContainerInterface $container
+     * @param string $requestedName
+     * @param array|null $options
      * @return IsGranted
      */
-    public function createService(ServiceLocatorInterface $serviceLocator)
+    public function __invoke(ContainerInterface $container, $requestedName, array $options = null)
     {
         /* @var \ZfcRbac\Service\AuthorizationService $authorizationService */
-        $authorizationService = $serviceLocator->getServiceLocator()->get('ZfcRbac\Service\AuthorizationService');
```
svycka: BC break?

src/ZfcRbac/Factory/IsGrantedViewHelperFactory.php
```diff
@@ -18,6 +18,10 @@
 
 namespace ZfcRbac\Factory;
 
+use Interop\Container\ContainerInterface;
+use Interop\Container\Exception\ContainerException;
```
svycka: not used

src/ZfcRbac/Factory/IsGrantedViewHelperFactory.php
```diff
@@ -18,6 +18,10 @@
...
+use Zend\ServiceManager\Exception\ServiceNotCreatedException;
```
svycka: not used

src/ZfcRbac/Factory/IsGrantedViewHelperFactory.php
```diff
@@ -18,6 +18,10 @@
...
+use Zend\ServiceManager\Exception\ServiceNotFoundException;
```
svycka: not used

src/ZfcRbac/Factory/ObjectRepositoryRoleProviderFactory.php
```diff
@@ -40,42 +39,55 @@ class ObjectRepositoryRoleProviderFactory implements FactoryInterface, MutableCr
...
     {
-        $parentLocator    = $serviceLocator->getServiceLocator();
+        if (! method_exists($container, 'build')) { // servicemanager v3
```
svycka: can be moved to `createService()`

src/ZfcRbac/Factory/ControllerGuardFactory.php
```diff
@@ -29,7 +28,7 @@
  * @author  Michaël Gallego <mic.gallego@gmail.com>
  * @license MIT
  */
-class ControllerGuardFactory implements FactoryInterface, MutableCreationOptionsInterface
```
svycka: removing `MutableCreationOptionsInterface` please look how servicemanager v2 keep compatibility: https://github.com/zendframework/zend-servicemanager/blob/release-2.7/src/AbstractPluginManager.php#L322-L328

basicaly instead of `MutableCreationOptionsInterface` it checks for `setCreationOptions()` not sure it works without it. Possible BC break unless this is handles somewhere else but don't know if this is good idea.

src/ZfcRbac/Factory/ObjectRepositoryRoleProviderFactory.php
```diff
@@ -40,42 +39,55 @@ class ObjectRepositoryRoleProviderFactory implements FactoryInterface, MutableCr
     /**
      * {@inheritDoc}
      */
-    public function setCreationOptions(array $options)
```
svycka: does it work with SMv2? if not see above

src/ZfcRbac/Factory/RouteGuardFactory.php
```diff
@@ -39,28 +38,40 @@ class RouteGuardFactory implements FactoryInterface, MutableCreationOptionsInter
     /**
      * {@inheritDoc}
      */
-    public function setCreationOptions(array $options)
```
svycka: BC break?

src/ZfcRbac/Factory/RoutePermissionsGuardFactory.php
```diff
@@ -41,28 +40,39 @@ class RoutePermissionsGuardFactory implements FactoryInterface, MutableCreationO
     /**
      * {@inheritDoc}
      */
-    public function setCreationOptions(array $options)
```
svycka: BC break?

src/ZfcRbac/Factory/RoutePermissionsGuardFactory.php
```diff
@@ -41,28 +40,39 @@ class RoutePermissionsGuardFactory implements FactoryInterface, MutableCreationO
...
-     * @param \Zend\ServiceManager\AbstractPluginManager|ServiceLocatorInterface $serviceLocator
-     * @return RouteGuard
+     * @param ContainerInterface $container
+     * @return RoutePermissionsGuard
      */
-    public function createService(ServiceLocatorInterface $serviceLocator)
+    public function __invoke(ContainerInterface $container)
     {
-        $parentLocator = $serviceLocator->getServiceLocator();
+        if (! method_exists($container, 'build')) { // servicemanager v3
```
svycka: can be moved to `createService()`

tests/ZfcRbacTest/Factory/RouteGuardFactoryTest.php
```diff
@@ -50,11 +50,15 @@ public function testFactory()
...
 
-        $pluginManager = new GuardPluginManager();
-        $pluginManager->setServiceLocator($serviceManager);
+        $pluginManager = new GuardPluginManager($serviceManager);
 
         $factory    = new RouteGuardFactory();
-        $routeGuard = $factory->createService($pluginManager);
+
+        if (method_exists($serviceManager, 'build')) { // servicemanager v3
+            $routeGuard = $factory->createService($serviceManager);
```
svycka: as others SMv3 does not call `createService()` so I don't know what you are testing here

tests/ZfcRbacTest/Factory/RoutePermissionsGuardFactoryTest.php
```diff
@@ -50,11 +50,16 @@ public function testFactory()
...
 
-        $pluginManager = new GuardPluginManager();
-        $pluginManager->setServiceLocator($serviceManager);
+        $pluginManager = new GuardPluginManager($serviceManager);
 
         $factory    = new RoutePermissionsGuardFactory();
-        $routeGuard = $factory->createService($pluginManager);
+
+        if (method_exists($serviceManager, 'build')) { // servicemanager v3
+            $routeGuard = $factory->createService($serviceManager);
```
svycka: as others SMv3 does not call createService() so I don't know what you are testing here

svycka:
@prolic Maybe didn't commented on all same problems I see, but you get the point. It is very hard to see if there is no other BC breaks as you modified old tests instead of creating new for SMv3. I would recommend keep old tests as much as possible, but maybe @bakura10 will test everything anyway :D

Maybe I am wrong in some cases because tests pass, but you changed tests in quite a lot of places so hard to tell. Guess you changed because they failed and that indicates BC break. Don't know why else you would have changed them.

src/ZfcRbac/Factory/ControllerPermissionsGuardFactory.php
```diff
@@ -18,19 +18,16 @@
...
 use ZfcRbac\Guard\ControllerPermissionsGuard;
-use ZfcRbac\Guard\RouteGuard;
 
 /**
  * Create a controller guard for checking permissions
  *
  * @author  JM Lerouxw <jmleroux.pro@gmail.com>
  * @license MIT
  */
-class ControllerPermissionsGuardFactory implements FactoryInterface, MutableCreationOptionsInterface
```
svycka: BC break? see https://github.com/zendframework/zend-servicemanager/pull/91#issuecomment-175772999

svycka:
forgot https://zendframework.github.io/zend-servicemanager/migration/#invokables those should be fixed also

tests/ZfcRbacTest/Factory/IsGrantedViewHelperFactoryTest.php
prolic: no, this is a change in zend service manager v3, the "main locator" instead of the plugin manager is passed

src/ZfcRbac/Factory/IsGrantedViewHelperFactory.php
prolic: same as above. ZF3 plugin manager send the main locator not the plugin manager itself to the factory

tests/ZfcRbacTest/Factory/ControllerGuardFactoryTest.php
prolic: createService is not used in v3

tests/ZfcRbacTest/Factory/IsGrantedViewHelperFactoryTest.php
svycka: @prolic it is passed to `__invoke()` not `createService()`

src/ZfcRbac/Factory/ObjectRepositoryRoleProviderFactory.php
prolic: you're right, i'll fix that tomorrow.

src/ZfcRbac/Factory/RouteGuardFactory.php
prolic: good idea.

src/ZfcRbac/Guard/AbstractGuard.php
prolic: why? self references the same class and is reusable when the class is renamed

src/ZfcRbac/Initializer/AuthorizationServiceInitializer.php
prolic: i'll check that tomorrow

src/ZfcRbac/Factory/ControllerGuardFactory.php
prolic: i'll re-check that tomorrow

prolic:
thans @svycka for your feedback. I'll try to improve most stuff tomorrow.

src/ZfcRbac/Guard/AbstractGuard.php
svycka: because it can be extended/changed if you leave it as is and also it was static before.

svycka: simple example:
```php
class foo extends bar
{
    const test = 1;

    public function test($test = self::test)
    {
        return $test;
    }
}
class bar extends foo
{
    const test = 2;
    public function test($test = self::test)
    {
        return $test;
    }
}
echo (new bar)->test(); // returns 2
echo (new foo)->test(); // returns 1
```

tests/ZfcRbacTest/Factory/ControllerGuardFactoryTest.php
svycka: yep you are right but you also added test to test call to it. Makes sense? no :)

composer.json
```diff
@@ -27,20 +27,27 @@
...
+    "zendframework/zend-serializer": "~2.2",
+    "zendframework/zend-view": "~2.2",
+    "phpunit/phpunit": "~4.8.26",
+    "squizlabs/php_codesniffer": "2.6.*",
     "satooshi/php-coveralls": "~0.6",
     "doctrine/common": "~2.4",
-    "doctrine/doctrine-module": "~0.8",
-    "doctrine/doctrine-orm-module": "~0.8"
+    "doctrine/doctrine-module": "~1.1",
+    "doctrine/doctrine-orm-module": "~0.11"
```
svycka: we already have stable version so "^1.0"

src/ZfcRbac/Initializer/AuthorizationServiceInitializer.php
```diff
@@ -32,17 +32,22 @@
...
             $instance->setAuthorizationService($authorizationService);
         }
     }
+
+    /**
+     * @see \Zend\ServiceManager\InitializerInterface::initialize()
+     */
+    public function initialize($instance, ServiceLocatorInterface $serviceLocator)
+    {
+        $this($serviceLocator, $instance);
```
svycka: should check if this is `AbstractPluginManager` and use `$serviceLocator->getServiceLocator()`

tests/ZfcRbacTest/Factory/AuthorizationServiceDelegatorTest.php
```diff
@@ -73,6 +74,10 @@ public function testAuthorizationServiceIsInjectedWithDelegator()
     {
         $serviceManager = ServiceManagerFactory::getServiceManager();
 
+        if (method_exists($serviceManager, 'build')) {
```
svycka: why do we need this?

tests/ZfcRbacTest/Factory/AuthorizationServiceDelegatorTest.php
```diff
@@ -95,6 +100,36 @@ public function testAuthorizationServiceIsInjectedWithDelegator()
         $this->assertEquals($authorizationService, $decoratedInstance->getAuthorizationService());
     }
 
+    public function testAuthorizationServiceIsInjectedWithDelegatorV3()
+    {
+        $serviceManager = ServiceManagerFactory::getServiceManager();
+
+        if (! method_exists($serviceManager, 'build')) {
```
svycka: why do we need this?

tests/ZfcRbacTest/Factory/AuthorizationServiceDelegatorTest.php
```diff
@@ -95,6 +100,36 @@ public function testAuthorizationServiceIsInjectedWithDelegator()
         $this->assertEquals($authorizationService, $decoratedInstance->getAuthorizationService());
     }
 
+    public function testAuthorizationServiceIsInjectedWithDelegatorV3()
```
svycka: maybe I don't see but where is the difference with `testAuthorizationServiceIsInjectedWithDelegator()` test?

tests/ZfcRbacTest/Factory/AuthorizationServiceDelegatorTest.php
```diff
@@ -119,10 +154,17 @@ function () {
...
 
-        $this->setExpectedException(
-            'ZfcRbac\Exception\RuntimeException',
-            'The service ZfcRbacTest\AuthorizationAware must implement AuthorizationServiceAwareInterface.'
-        );
-        $serviceManager->get('ZfcRbacTest\AuthorizationAware');
+        $thrown = false;
+        try {
+            $serviceManager->get('ZfcRbacTest\AuthorizationAware');
+        } catch (\Exception $e) {
```
svycka: why you changed this? does this means that we have a BC break?

tests/ZfcRbacTest/Factory/AuthorizationServiceDelegatorTest.php
prolic: this is to check whether servicemanager v2 oder v3 is used.
i would have used https://github.com/Ocramius/PackageVersions, but this is php7 only.

tests/ZfcRbacTest/Factory/AuthorizationServiceDelegatorTest.php
prolic: the difference is

```
$serviceManager->addDelegator(
    'ZfcRbacTest\Initializer\AuthorizationAwareFake',
    'ZfcRbac\Factory\AuthorizationServiceDelegatorFactory'
);
```
VS 
```
$serviceManager->addDelegator(
    'ZfcRbacTest\AuthorizationAware',
    'ZfcRbac\Factory\AuthorizationServiceDelegatorFactory'
);
```

tests/ZfcRbacTest/Factory/AuthorizationServiceDelegatorTest.php
prolic: the `ZfcRbac\Exception\RuntimeException` is not thrown, it's only the  previous exception in zf3, because a ServiceNotCreatedException is thrown in zf3.

tests/ZfcRbacTest/Factory/ControllerPermissionsGuardFactoryTest.php
```diff
@@ -43,20 +49,51 @@ public function testFactory()
...
-        $serviceManager = new ServiceManager();
         $serviceManager->setService('ZfcRbac\Options\ModuleOptions', $options);
         $serviceManager->setService(
             'ZfcRbac\Service\AuthorizationService',
             $this->getMock('ZfcRbac\Service\AuthorizationService', [], [], '', false)
         );
 
-        $pluginManager = new GuardPluginManager();
-        $pluginManager->setServiceLocator($serviceManager);
+        $pluginManager = new GuardPluginManager($serviceManager);
```
svycka: why did you changed this should have worked without changes with SMv2

prolic: yes it would, but as i said, the factories have BC break, if you extended or aggregated them, you might need to update your code anyway. this way we can ensure the "new way" of injecting via constructor is used everywhere.

tests/ZfcRbacTest/Initializer/AuthorizationServiceInitializerTest.php
```diff
@@ -31,15 +33,16 @@ public function testInitializer()
...
-            ->method('get')
-            ->with($authServiceClassName)
-            ->will($this->returnValue($authorizationService));
+        $serviceLocator
+            ->get($authServiceClassName)
+            ->willReturn($authorizationService)
+            ->shouldBeCalled();
 
-        $initializer->initialize($instance, $serviceLocator);
+        $initializer->initialize($instance, $serviceLocator->reveal());
```
svycka: do you test SMv2 or v3 here?

svycka:
@prolic maybe we could use `Zend\ServiceManager\Test\CommonPluginManagerTrait`  more: https://zendframework.github.io/zend-servicemanager/migration/#migration-testing for plugin manager tests

prolic:
We could make use of `Zend\ServiceManager\Test\CommonPluginManagerTrait`, i cannot do it today anymore. And it's only a single test that makes this problem and I solved it another way. We can also leave it as is I think.

tests/ZfcRbacTest/Initializer/AuthorizationServiceInitializerTest.php
prolic: v3 service locator interface exents interop config interface, that's why I switched to prophecy here to be able to have a mock implements 2 interfaces.

svycka: @prolic I am about the fact that $serviceLocator can be pluginManager and you removed check for this in the `AuthorizationServiceInitializer`

prolic: this is re-added here: https://github.com/ZF-Commons/zfc-rbac/pull/335/commits/4dc3eb097b4a009c841d82aa3a1fb5b2f8065959

prolic:
Actually making use of CommonPluginManagerTrait makes only sense, when the plugin manager / factory tests are changed.

The "old" tests are like `$factory->createService($sm / $pluginManager);` and check whether it's successful and correct. What we need to have a test for v2 and v3 would be:

`$service = $pluginManager->get('some service name');`, that would make the tests for v2 and v3 consistent (no special tests needed). On the other hand the factories have BC break and when you extended or aggregated them, you may need to make adjustments anyway. There no way to go around this, so I see no reason to invest further time in this.

nasko:
Not sure if this is the proper place to discuss this, but I've encountered a problem and because in my view this coin has two sides too, bringing it up here is just one of my options.

I'm using:

- https://github.com/zendframework/zend-mvc/commit/d039a45a167e18323b874da396f5806fbeca04aa
  -  this is the most recent commit of the develop branch
- https://github.com/zendframework/zend-eventmanager/commit/28b26fff230ffb8976d1d71fd1305ce886326d86
  - this is the most recent commit too
- https://github.com/prolic/zfc-rbac/commit/4dc3eb097b4a009c841d82aa3a1fb5b2f8065959
  - this is the last but one commit in the zf3 branch

When the `MvcEvent::EVENT_DISPATCH_ERROR` is triggered in the [AbstractGuard:onResult()](https://github.com/prolic/zfc-rbac/blob/4dc3eb097b4a009c841d82aa3a1fb5b2f8065959/src/ZfcRbac/Guard/AbstractGuard.php#L76) method, all registered listeners receive a `Zend\Mvc\MvcEvent` event as their first argument.

However, when the event is handled in [Zend\EventManager\EventManager::trigger()](https://github.com/zendframework/zend-eventmanager/blob/28b26fff230ffb8976d1d71fd1305ce886326d86/src/EventManager.php#L115), the event is transformed into a `Zend\EventManager\Event` instance:

```php
public function trigger($eventName, $target = null, $argv = [])
{
    $event = clone $this->eventPrototype; // Zend\EventManager\Event
    $event->setName($eventName);
    $event->setTarget($target);
    $event->setParams($argv);

    return $this->triggerListeners($event);
}
```
However, all registered listeners expect a `Zend\Mvc\MvcEvent` instance, but receive `Zend\EventManager\Event`, which causes fatal errors:

```
Catchable fatal error: Argument 1 passed to Zend\Mvc\View\Http\RouteNotFoundStrategy::detectNotFoundError() must be an instance of Zend\Mvc\MvcEvent, instance of Zend\EventManager\Event given, called in /var/www/projects/***/vendor/zendframework/zend-eventmanager/src/EventManager.php on line 271 and defined in /var/www/projects/***/vendor/zendframework/zend-mvc/src/View/Http/RouteNotFoundStrategy.php on line 135
```

Now, `RouteNotFoundStrategy::detectNotFoundError()` is just the first of a queue of 5 listeners, `ZfcRbac\View\Strategy\RedirectStrategy` being the last but one in the list.

Noticing this, I decided to simulate a request with faulty route - I requested http://my-applciation.dev/blabla, just to trace how the `dispatch.error` would be handled and how the above mentioned `RouteNotFoundStrategy::detectNotFoundError()` listener would handle the wrong type of the event instance. But, surprisingly the event passed to the listener was of the correct `MvcEvent` type.
Digging into the trace, I noticed that the event was triggered by `Zend\EventManager\EventManager::triggerEvent()` and not by `::trigger()`.

Here's the relevant section of [Zend\Mvc\RouteListener](https://github.com/zendframework/zend-mvc/blob/d039a45a167e18323b874da396f5806fbeca04aa/src/RouteListener.php#L56):

```php
public function onRoute(MvcEvent $event)
{
    // ...
    $target  = $event->getTarget();
    $results = $target->getEventManager()->triggerEvent($event);
    if (!empty($results)) {
        return $results->last();
    }

    return $event->getParams();
}
```

At first glance, the difference between `EventManager::trigger()` and `EventManager::triggerEvent()` is that the former transforms the type of the event instance and the latter merely triggers the registered listeners.

Looking at the history of the `RouteListener` class, I noticed that its implementation was changed recently with [this commit](https://github.com/zendframework/zend-mvc/commit/3bdb9f41d3bb589769b2570589728eaca941af67#diff-585cd82140bc763521c4d4a02b41a460):

```
@@ -44,9 +45,10 @@ public function onRoute($e)
          $routeMatch = $router->match($request);
  
          if (!$routeMatch instanceof Router\RouteMatch) {
 +            $e->setName(MvcEvent::EVENT_DISPATCH_ERROR);
              $e->setError(Application::ERROR_ROUTER_NO_MATCH);
  
 -            $results = $target->getEventManager()->trigger(MvcEvent::EVENT_DISPATCH_ERROR, $e);
 +            $results = $target->getEventManager()->triggerEvent($e);
              if (count($results)) {
                  return $results->last();
              }
```

Please forgive my long introduction, but my question is: should ZfcRbac use `EventManager::triggerEvent()` instead of `trigger()`, to address the apparently changed eventmanager implementation and additionally not to cause for a wrong event instance type to be passed over?

I browsed the ZfcRbac tests and obviously they pass, because a mock implementation of EventManager is being used and in addition, in the tests the listener is called directly:

```
$redirectStrategy->onError($mvcEvent);
```

nasko:
Not sure if this is the proper place to discuss this, but I've encountered a problem and because in my view this coin has two sides too, bringing it up here is just one of my options.

I'm using:

- https://github.com/zendframework/zend-mvc/commit/d039a45a167e18323b874da396f5806fbeca04aa
  -  this is the most recent commit of the develop branch
- https://github.com/zendframework/zend-eventmanager/commit/28b26fff230ffb8976d1d71fd1305ce886326d86
  - this is the most recent commit too
- https://github.com/prolic/zfc-rbac/commit/4dc3eb097b4a009c841d82aa3a1fb5b2f8065959
  - this is the last but one commit in the zf3 branch

When the `MvcEvent::EVENT_DISPATCH_ERROR` is triggered in the [AbstractGuard:onResult()](https://github.com/prolic/zfc-rbac/blob/4dc3eb097b4a009c841d82aa3a1fb5b2f8065959/src/ZfcRbac/Guard/AbstractGuard.php#L76) method, all registered listeners receive a `Zend\Mvc\MvcEvent` event as their first argument.

However, when the event is handled in [Zend\EventManager\EventManager::trigger()](https://github.com/zendframework/zend-eventmanager/blob/28b26fff230ffb8976d1d71fd1305ce886326d86/src/EventManager.php#L115), the event is transformed into a `Zend\EventManager\Event` instance:

```php
public function trigger($eventName, $target = null, $argv = [])
{
    $event = clone $this->eventPrototype; // Zend\EventManager\Event
    $event->setName($eventName);
    $event->setTarget($target);
    $event->setParams($argv);

    return $this->triggerListeners($event);
}
```
However, all registered listeners expect a `Zend\Mvc\MvcEvent` instance, but receive `Zend\EventManager\Event`, which causes fatal errors:

```
Catchable fatal error: Argument 1 passed to Zend\Mvc\View\Http\RouteNotFoundStrategy::detectNotFoundError() must be an instance of Zend\Mvc\MvcEvent, instance of Zend\EventManager\Event given, called in /var/www/projects/***/vendor/zendframework/zend-eventmanager/src/EventManager.php on line 271 and defined in /var/www/projects/***/vendor/zendframework/zend-mvc/src/View/Http/RouteNotFoundStrategy.php on line 135
```

Now, `RouteNotFoundStrategy::detectNotFoundError()` is just the first of a queue of 5 listeners, `ZfcRbac\View\Strategy\RedirectStrategy` being the last but one in the list.

Noticing this, I decided to simulate a request with faulty route - I requested http://my-applciation.dev/blabla, just to trace how the `dispatch.error` would be handled and how the above mentioned `RouteNotFoundStrategy::detectNotFoundError()` listener would handle the wrong type of the event instance. But, surprisingly the event passed to the listener was of the correct `MvcEvent` type.
Digging into the trace, I noticed that the event was triggered by `Zend\EventManager\EventManager::triggerEvent()` and not by `::trigger()`.

Here's the relevant section of [Zend\Mvc\RouteListener](https://github.com/zendframework/zend-mvc/blob/d039a45a167e18323b874da396f5806fbeca04aa/src/RouteListener.php#L56):

```php
public function onRoute(MvcEvent $event)
{
    // ...
    $target  = $event->getTarget();
    $results = $target->getEventManager()->triggerEvent($event);
    if (!empty($results)) {
        return $results->last();
    }

    return $event->getParams();
}
```

At first glance, the difference between `EventManager::trigger()` and `EventManager::triggerEvent()` is that the former transforms the type of the event instance and the latter merely triggers the registered listeners.

Looking at the history of the `RouteListener` class, I noticed that its implementation was changed recently with [this commit](https://github.com/zendframework/zend-mvc/commit/3bdb9f41d3bb589769b2570589728eaca941af67#diff-585cd82140bc763521c4d4a02b41a460):

```
@@ -44,9 +45,10 @@ public function onRoute($e)
          $routeMatch = $router->match($request);
  
          if (!$routeMatch instanceof Router\RouteMatch) {
 +            $e->setName(MvcEvent::EVENT_DISPATCH_ERROR);
              $e->setError(Application::ERROR_ROUTER_NO_MATCH);
  
 -            $results = $target->getEventManager()->trigger(MvcEvent::EVENT_DISPATCH_ERROR, $e);
 +            $results = $target->getEventManager()->triggerEvent($e);
              if (count($results)) {
                  return $results->last();
              }
```

Please forgive my long introduction, but my question is: should ZfcRbac use `EventManager::triggerEvent()` instead of `trigger()`, to address the apparently changed eventmanager implementation and additionally not to cause for a wrong event instance type to be passed over?

I browsed the ZfcRbac tests and obviously they pass, because a mock implementation of EventManager is being used and in addition, in the tests the listener is called directly:

```
$redirectStrategy->onError($mvcEvent);
```

nasko:
Not sure if this is the proper place to discuss this, but I've encountered a problem and because in my view this coin has two sides too, bringing it up here is just one of my options.

I'm using:

- https://github.com/zendframework/zend-mvc/commit/d039a45a167e18323b874da396f5806fbeca04aa
  -  this is the most recent commit of the develop branch
- https://github.com/zendframework/zend-eventmanager/commit/28b26fff230ffb8976d1d71fd1305ce886326d86
  - this is the most recent commit too
- https://github.com/prolic/zfc-rbac/commit/4dc3eb097b4a009c841d82aa3a1fb5b2f8065959
  - this is the last but one commit in the zf3 branch

When the `MvcEvent::EVENT_DISPATCH_ERROR` is triggered in the [AbstractGuard:onResult()](https://github.com/prolic/zfc-rbac/blob/4dc3eb097b4a009c841d82aa3a1fb5b2f8065959/src/ZfcRbac/Guard/AbstractGuard.php#L76) method, all registered listeners receive a `Zend\Mvc\MvcEvent` event as their first argument.

However, when the event is handled in [Zend\EventManager\EventManager::trigger()](https://github.com/zendframework/zend-eventmanager/blob/28b26fff230ffb8976d1d71fd1305ce886326d86/src/EventManager.php#L115), the event is transformed into a `Zend\EventManager\Event` instance:

```php
public function trigger($eventName, $target = null, $argv = [])
{
    $event = clone $this->eventPrototype; // Zend\EventManager\Event
    $event->setName($eventName);
    $event->setTarget($target);
    $event->setParams($argv);

    return $this->triggerListeners($event);
}
```
However, all registered listeners expect a `Zend\Mvc\MvcEvent` instance, but receive `Zend\EventManager\Event`, which causes fatal errors:

```
Catchable fatal error: Argument 1 passed to Zend\Mvc\View\Http\RouteNotFoundStrategy::detectNotFoundError() must be an instance of Zend\Mvc\MvcEvent, instance of Zend\EventManager\Event given, called in /var/www/projects/***/vendor/zendframework/zend-eventmanager/src/EventManager.php on line 271 and defined in /var/www/projects/***/vendor/zendframework/zend-mvc/src/View/Http/RouteNotFoundStrategy.php on line 135
```

Now, `RouteNotFoundStrategy::detectNotFoundError()` is just the first of a queue of 5 listeners, `ZfcRbac\View\Strategy\RedirectStrategy` being the last but one in the list.

Noticing this, I decided to simulate a request with faulty route - I requested http://my-applciation.dev/blabla, just to trace how the `dispatch.error` would be handled and how the above mentioned `RouteNotFoundStrategy::detectNotFoundError()` listener would handle the wrong type of the event instance. But, surprisingly the event passed to the listener was of the correct `MvcEvent` type.
Digging into the trace, I noticed that the event was triggered by `Zend\EventManager\EventManager::triggerEvent()` and not by `::trigger()`.

Here's the relevant section of [Zend\Mvc\RouteListener](https://github.com/zendframework/zend-mvc/blob/d039a45a167e18323b874da396f5806fbeca04aa/src/RouteListener.php#L56):

```php
public function onRoute(MvcEvent $event)
{
    // ...
    $target  = $event->getTarget();
    $results = $target->getEventManager()->triggerEvent($event);
    if (!empty($results)) {
        return $results->last();
    }

    return $event->getParams();
}
```

At first glance, the difference between `EventManager::trigger()` and `EventManager::triggerEvent()` is that the former transforms the type of the event instance and the latter merely triggers the registered listeners.

Looking at the history of the `RouteListener` class, I noticed that its implementation was changed recently with [this commit](https://github.com/zendframework/zend-mvc/commit/3bdb9f41d3bb589769b2570589728eaca941af67#diff-585cd82140bc763521c4d4a02b41a460):

```
@@ -44,9 +45,10 @@ public function onRoute($e)
          $routeMatch = $router->match($request);
  
          if (!$routeMatch instanceof Router\RouteMatch) {
 +            $e->setName(MvcEvent::EVENT_DISPATCH_ERROR);
              $e->setError(Application::ERROR_ROUTER_NO_MATCH);
  
 -            $results = $target->getEventManager()->trigger(MvcEvent::EVENT_DISPATCH_ERROR, $e);
 +            $results = $target->getEventManager()->triggerEvent($e);
              if (count($results)) {
                  return $results->last();
              }
```

Please forgive my long introduction, but my question is: should ZfcRbac use `EventManager::triggerEvent()` instead of `trigger()`, to address the apparently changed eventmanager implementation and additionally not to cause for a wrong event instance type to be passed over?

I browsed the ZfcRbac tests and obviously they pass, because a mock implementation of EventManager is being used and in addition, in the tests the listener is called directly:

```
$redirectStrategy->onError($mvcEvent);
```

prolic:
@nasko Submit a PR to my zf3 branch and add some unit tests (failing without your patch).

nasko:
> @nasko Submit a PR to my zf3 branch and add some unit tests (failing without your patch).

I'm not sure if I can do this properly. Isn't the purpose of the unit tests to test implementation in isolation, while the issue I've encountered is manifested when used in conjunction with zend-mvc and zend-eventmanager, which both have been changed recently?

prolic:
the test should be failing with current implementation and pass with your
patch. it should work with zf2 and zf3, then we are safe.

Sascha-Oliver Prolic

2016-07-20 22:04 GMT+08:00 Atanas Vasilev <notifications@github.com>:

nasko:
I could really use some help here.

In my situation, if I change this:

```php
    public function onResult(MvcEvent $event)
    {
        if ($this->isGranted($event)) {
            return;
        }
        $event->setError(self::GUARD_UNAUTHORIZED);
        $event->setParam('exception', new Exception\UnauthorizedException(
            'You are not authorized to access this resource',
            403
        ));
        $event->stopPropagation(true);
        $application  = $event->getApplication();
        $eventManager = $application->getEventManager();
        $eventManager->trigger(MvcEvent::EVENT_DISPATCH_ERROR, $event);
    }

```

... to this:

```php
    public function onResult(MvcEvent $event)
    {
        if ($this->isGranted($event)) {
            return;
        }

        $event->setError(self::GUARD_UNAUTHORIZED);
        $event->setParam('exception', new Exception\UnauthorizedException(
            'You are not authorized to access this resource',
            403
        ));

        $event->stopPropagation(true);
        $event->setName(MvcEvent::EVENT_DISPATCH_ERROR);

        $target  = $event->getTarget();
        $target->getEventManager()->triggerEvent($event);
    }
```

The fatal error is not emitted and the redirect strategy does its job properly.

What I don't know is, how I could represent this with unit tests. This is the test I'm looking at: 

`ControllerGuardTest::testProperlySetUnauthorizedAndTriggerEventOnUnauthorization()`

```php
    public function testProperlySetUnauthorizedAndTriggerEventOnUnauthorization()
    {
        $event      = new MvcEvent();
        $routeMatch = new RouteMatch([]);

        $application  = $this->getMock('Zend\Mvc\Application', [], [], '', false);
        $eventManager = $this->getMock('Zend\EventManager\EventManagerInterface');

        $application->expects($this->once())
                    ->method('getEventManager')
                    ->will($this->returnValue($eventManager));

        $eventManager->expects($this->once())
                     ->method('trigger')
                     ->with(MvcEvent::EVENT_DISPATCH_ERROR);

        $routeMatch->setParam('controller', 'MyController');
        $routeMatch->setParam('action', 'delete');

        $event->setRouteMatch($routeMatch);
        $event->setApplication($application);

        $identityProvider = $this->getMock('ZfcRbac\Identity\IdentityProviderInterface');
        $identityProvider->expects($this->any())
                         ->method('getIdentityRoles')
                         ->will($this->returnValue('member'));

        $roleProvider = new InMemoryRoleProvider([
            'member'
        ]);

        $roleService = new RoleService($identityProvider, $roleProvider, new RecursiveRoleIteratorStrategy());

        $routeGuard = new ControllerGuard($roleService, [[
            'controller' => 'MyController',
            'actions'    => 'edit',
            'roles'      => 'member'
        ]]);

        $routeGuard->onResult($event);

        $this->assertTrue($event->propagationIsStopped());
        $this->assertEquals(ControllerGuard::GUARD_UNAUTHORIZED, $event->getError());
        $this->assertInstanceOf('ZfcRbac\Exception\UnauthorizedException', $event->getParam('exception'));
    }
```

You see, the event propagation, error and exception param are set okay, but when `$eventManager->trigger(MvcEvent::EVENT_DISPATCH_ERROR, $event);` is called, the event is passed further as the target and there the registered listeners are notified with the wrong event type:

```php
namespace Zend\EventManager;

class EventManager implements EventManagerInterface
{
    public function trigger($eventName, $target = null, $argv = [])
    {
        // Zend\EventManager\Event and not Zend\Mvc\MvcEvent 
        // which all listeners expect
        $event = clone $this->eventPrototype; 
        $event->setName($eventName);
        $event->setTarget($target);
        $event->setParams($argv);

        return $this->triggerListeners($event);
    }
}
```

I don't know how to test this in isolation as in my opinion it's an integration issue. Could someone advise as to how this could be handled?

bupy7:
When planned merges these changes?

nasko:
> it should work with zf2 and zf3, then we are safe

@prolic, How should I implement the tests to pass both with zf2 and zf3, with respect to changed namespaces? For example in ZF3 the Router is in its own namespace `Zend\Router` vs. `Zend\Mvc\Router` which is used in the current tests. 
The tests won't pass on my environment, unless I edit the imports that reference the old namespace `Zend\Mvc\Router`

webimpress:
@nasko maybe it will help:
https://github.com/zfcampus/zf-content-validation/blob/master/test/ContentValidationListenerTest.php#L17-L34

nasko:
@webimpress Thank you! The example is valid, but the same approach is not used with the current tests in the `zf3` branch.

The tests fail with my environment even without modifying the code.

nasko:
Okay, I'm not creating a PR, because of said differences regarding imports, but here's my commit that basically fixes the problem for me: https://github.com/nasko/zfc-rbac/commit/9e97ce306f4b3d1fbdcc3a3db416523446e59e95.

I'd be grateful for any feedback.

danizord:
@nasko you can skip the test if class_exists returns false.

danizord:
I'm going do merge it tonight, @bakura10 can you review it as well?

nasko:
This is beyond me - I guess I'll have to resort to using my fork to address #340

prolic:
updated, thanks to @nasko for helping to fix this.

src/ZfcRbac/Factory/RoutePermissionsGuardFactory.php
```diff
@@ -41,28 +41,45 @@ class RoutePermissionsGuardFactory implements FactoryInterface, MutableCreationO
...
-
         /* @var \ZfcRbac\Options\ModuleOptions $moduleOptions */
-        $moduleOptions = $parentLocator->get('ZfcRbac\Options\ModuleOptions');
+        $moduleOptions = $container->get('ZfcRbac\Options\ModuleOptions');
 
         /* @var \ZfcRbac\Service\AuthorizationService $authorizationService */
-        $authorizationService = $parentLocator->get('ZfcRbac\Service\AuthorizationService');
+        $authorizationService = $container->get('ZfcRbac\Service\AuthorizationService');
 
         $routeGuard = new RoutePermissionsGuard($authorizationService, $this->options);
```
danizord: `$options`?

prolic: fixed.

svycka:
@danizord @bakura10 can we have a release with this?

PR #342, opened 2016-09-04
Change/remove auth service from assertions

basz:
This PR removes the AuthorizationService as argument to Assertions.

This mean you now should use a factory to instantiate assertions, but only when you actually use the AuthorizationService.


bakura10:
@danizord thoughts on that?

danizord:
LGFM :+1:

PR #343, opened 2016-09-04
Change/pass context parameter to RoleService::getIdentityRoles

basz:
This PR will pass the context to the RoleService::getIdentityRoles from within the AuthorizationService. Additionally it adds an RoleServiceInterface so the RoleService can be changed.

Not a breaking change.

As requested by @Wilt in https://github.com/ZF-Commons/zfc-rbac/pull/316

### DIFF
--- a/src/Service/RoleServiceInterface.php
+++ b/src/Service/RoleServiceInterface.php
@@ -1,0 +1,40 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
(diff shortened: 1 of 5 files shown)


src/Service/RoleServiceInterface.php
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * @author  Michaël Gallego <mic.gallego@gmail.com>
+ * @licence MIT
+ */
+interface RoleServiceInterface
+{
+    /**
+     * Get the identity roles from the current identity, applying some more logic
+     *
+     * @param IdentityInterface $identity
```
danizord: null|IdentityInterface

src/Service/RoleServiceInterface.php
```diff
@@ -0,0 +1,40 @@
...
+     * @param null              $context
```
danizord: mixed

PR #344, opened 2016-09-04
Change/pass permission as param to assertions

basz:
As proposed this PR passes the requested permission name to the Assertion::assert method, so the assertion can assert based on that information.

Note: based on an PR #342 that is also a proposal. To avoid merge conflicts please review/merge that one first.

### DIFF
--- a/src/Service/AuthorizationService.php
+++ b/src/Service/AuthorizationService.php
@@ -117,7 +117,7 @@ class AuthorizationService implements AuthorizationServiceInterface
         }
 
         if ($this->hasAssertion($permission)) {
-            return $this->assert($this->assertions[(string) $permission], $identity, $context);
+            return $this->assert($this->assertions[(string) $permission], $permission, $identity, $context);
         }
 
         return true;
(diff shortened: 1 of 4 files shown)


danizord:
LGFM :+1:

prolic:
I don't see a use-case. Can someone explain it to me?

basz:
```
public function assert($permission, IdentityInterface $identity = null, $context = null): bool {
    if ((string) $identity->getUserId() === (string) $context->userId()) {
        return true;
    }

    return $this->authorizationService->isGranted($identity, sprint("%s.forOthers", $permission, $context));
}
```

prolic:
ok, good for me then!
:+1:

basz:
rebased

src/Service/AuthorizationService.php
```diff
@@ -117,33 +117,33 @@ public function isGranted($identity, $permission, $context = null)
         }
 
         if ($this->hasAssertion($permission)) {
-            return $this->assert($this->assertions[(string) $permission], $identity, $context);
+            return $this->assert($this->assertions[(string) $permission], $permission, $identity, $context);
```
bakura10: Should you cast to string as well as you're saying that in the docblock? Otherwise you may receive an AssertionInterface.

basz: I see what you mean, but why/when would a permission be an AssertionInterface? I would argue never so the cast can be removed.

bakura10: Actually we have this interface (https://github.com/ZF-Commons/zfc-rbac/blob/develop/src/Assertion/AssertionInterface.php) and it does not implement __toString.

Maybe we should remove this interface completely?

basz: I think we should keep that.

$permission is not $this->assertions[(string) $permission]

And permission is always a string, is is not?

basz: either you or I am confused :-p

bakura10: I must be the one who is confused :p. I haven't use it for too much time.

basz: :-) can I do anything to help you get unconfused?

bakura10: Never mind, I've consued assertion and permission... ✌️

Issue #348, opened 2016-10-06
Tag stable version for ZF3 compat


exptom:
master currently supports ZF3 and I am using it in a project successfully. Is it possible to get it tagged so that I don't have to rely on a dev version in composer.json?

Thanks :)

bakura10:
Hi,

Thanks. I'm thinking of tagging master as 2.6.0 (version 3 still in development is the develop branch). @danizord any objection on that?

cincarnato:
+1

bakura10:
I've just tried to pull the latest code and run the tests and I'm having +67 errors, apparently due to module manager, which is a bit strange...

I'll therefore need to delay that for now to dig into the issue. If someone can also try to pull the dev-master, and try to run the tests to see what is happening that would help as well.

@danizord are you able to reproduce this issue?

webimpress:
@bakura10 please see my PR #349

bakura10:
I'm happy to say that ZfcRbac has been tagged to 2.6.0 :). Thanks a lot!

I will open another issue in a few seconds for looking for a new leader for this project!

exptom:
Thank you :)

PR #349, opened 2016-10-14
Tests fixes - pass on ZF2 and ZF3

webimpress:
As noted in https://github.com/ZF-Commons/zfc-rbac/issues/348#issuecomment-253732132 there were some errors in tests. I've fixed it to pass on ZF2 and ZF3 modules.

I'd prefer also to update travis configuration, to have something similar as we have in ZF/Apigility modules:
https://github.com/zendframework/zend-view/blob/master/.travis.yml
or
https://github.com/zfcampus/zf-hal/blob/master/.travis.yml


bakura10:
Thanks. This fixed the issue here.

PR #352, opened 2016-10-20
Fix #351 - Added missing import

### LINKED ISSUE #351: Missing use statement in ZfcRbac\Initializer\AuthorizationServiceInitializer
AuthorizationServiceInitializer @ Line 51 asks for $serviceLocator instanceof AbstractPluginManager
https://github.com/ZF-Commons/zfc-rbac/blob/master/src/ZfcRbac/Initializer/AuthorizationServiceInitializer.php#L51

but
use Zend\ServiceManager\AbstractPluginManager; 
is missing

webimpress:
Please see #351


bakura10:
Good catch! Thanks!

Issue #357, opened 2016-10-26
RedirectStrategy doesn't work with "ZF3" Eventmanager


DennisDobslaf:
I've installed version 2.6.3 of zendframework/zend-eventmanager (method '**triggerEvent**' is available => your condition in https://github.com/ZF-Commons/zfc-rbac/blob/master/src/ZfcRbac/Guard/AbstractGuard.php#L76). So ZfcRbac decides to go the "ZF3 EventManager way".

But the method isn't new in Version 3 of the EventManager!

In my case, this leads to a non executed RedirectStrategey, because "stopPropagation" is not set to false as a default and the eventmanager stops the execution right after the very first attached listener. 

Asking for a method to exist is a bad idea for checking the version in my eyes. Unfortunately zf eventmanager doesn't provide version information via the interface. 
Ignoring the fact, **setEventPrototype** seems to be a new (public) method in ZF3 EventManager.

Changing Line 76 to: 
`       if (method_exists($eventManager, 'setEventPrototype')) {`
works for me.

svycka:
hmm yep this whole check is not necessary could be just:
```php
$event->setName(MvcEvent::EVENT_DISPATCH_ERROR);
$eventManager->triggerEvent($event);
```
because this would work with EventManager v2 and v3

DennisDobslaf:
L71 should be 
`$event->stopPropagation(false);`
then.

Because v2 Eventmanager doesn't set this in **triggerEvent** but in **trigger** (which wouldn't be called anymore).

svycka:
I think this by default is `false` but maybe I am wrong.

svycka:
woops missed the point :D hmm just a quick look but this `$event->stopPropagation(true);` also not required haven't tested but why we need to stop? again missing something? :)

DennisDobslaf:
Off course, you are right. false is the default value.

So, changing the lines like your comment and deleting stopPropagation(true) works (for me) with v2. Currently i'm not able to validate this with v3.

This is the complete AbstractGuard::onResult
```
/**
 * @private
 * @param  MvcEvent $event
 * @return void
 */
public function onResult(MvcEvent $event)
{
    if ($this->isGranted($event)) {
        return;
    }

    $event->setError(self::GUARD_UNAUTHORIZED);
    $event->setParam('exception', new Exception\UnauthorizedException(
        'You are not authorized to access this resource',
        403
    ));

    $application  = $event->getApplication();
[...]

PR #358, opened 2016-10-26
stop propagation only after MvcEvent::EVENT_DISPATCH_ERROR

svycka:
because only one listener would be triggered
should fix https://github.com/ZF-Commons/zfc-rbac/issues/357


prolic:
Can you add a test-case for this, please?

src/ZfcRbac/Guard/AbstractGuard.php
```diff
@@ -68,18 +68,13 @@ public function onResult(MvcEvent $event)
...
-        if (method_exists($eventManager, 'triggerEvent')) {
-            // ZF3 EventManager
-            $event->setName(MvcEvent::EVENT_DISPATCH_ERROR);
-            $eventManager->triggerEvent($event);
-        } else {
-            // ZF2 EventManager
-            $eventManager->trigger(MvcEvent::EVENT_DISPATCH_ERROR, $event);
-        }
+        $event->setName(MvcEvent::EVENT_DISPATCH_ERROR);
+        $eventManager->triggerEvent($event);
```
svycka: just in case someone will come and complain about BC break. Yes this can be BC break because we depend on `EventManagerInterface` and if someone crazy enough implement it themselves and not use EventManager this is BC break but do we care? :)

prolic:
Write a test that fails without this patch and succeeds with this patch.

tests/ZfcRbacTest/Guard/RoutePermissionsGuardTest.php
```diff
@@ -410,7 +410,7 @@ public function testProperlyFillEventOnAuthorization()
 
     public function testProperlySetUnauthorizedAndTriggerEventOnUnauthorization()
     {
-        $eventManager = $this->getMock('Zend\EventManager\EventManagerInterface');
+        $eventManager = $this->getMock('Zend\EventManager\EventManager');
```
prolic: Two questions:
1) Why is this needed?
2) How is this a BC break, when it's only in the tests?

webimpress:
@svycka do we need still these conditions in tests:

```php
         if (method_exists($eventManager, 'triggerEvent')) {
             $eventManager->expects($this->once())
                          ->method('triggerEvent')
                          ->with($event);
         } else {
             $eventManager->expects($this->once())
                          ->method('trigger')
                          ->with(MvcEvent::EVENT_DISPATCH_ERROR);
         }
```
?

tests/ZfcRbacTest/Guard/RoutePermissionsGuardTest.php
prolic: Ah I see, because triggerEvent is not part of the 2.x series interface.

prolic:
@webimpress correct, this could also be updated.

tests/ZfcRbacTest/Guard/RoutePermissionsGuardTest.php
svycka: because `EventManagerInterface` does not have `triggerEvent()` but `EventManager` has it

svycka:
@prolic anything else?

prolic:
Thx

PR #359, opened 2016-10-27
Update ControllerPermissionsGuard add condition check

Itach1Uchixa:
ControllerPermissionsGuard add condition check


prolic:
Coverage decreased, some more tests?

Itach1Uchixa:
I don't understand what causes this error, I think it's because I added some new unit tests. Help me!

basz:
The reason you get an error is because you've added code... so on average the score when drown

Itach1Uchixa:
As I understand it is not a error or big problem, right?

basz:
Correct it is not an error, nor a 'big' problem.

However, we kindly request tests are written for all code that is added.

PR #361, opened 2017-01-26
Updated usage of Event attach method for ZF3

divix1988:


docs/02. Quick Start.md
```diff
@@ -93,10 +93,14 @@ the following code in the `onBootstrap` method of your `Module.php` class:
 ```php
 public function onBootstrap(EventInterface $e)
```
danizord: Change typehint to MvcEvent (since there's no `getApplication()` method in `EventInterface`)

docs/02. Quick Start.md
```diff
@@ -93,10 +93,14 @@ the following code in the `onBootstrap` method of your `Module.php` class:
...
 {
-    $t = $e->getTarget();
-
-    $t->getEventManager()->attach(
-        $t->getServiceManager()->get('ZfcRbac\View\Strategy\RedirectStrategy')
+    $app = $e->getApplication();
+    $sm = $app->getServiceManager();
+    
+    $e->getTarget()->getEventManager()->attach(
```
danizord: I think you can use `$app` here instead of `$e->getTarget()`

svycka: for zf3 it should be as follows:
```php
$listener = $sm->get('ZfcRbac\View\Strategy\RedirectStrategy');
$listener->attach($eventManager);
// or one liner
($sm->get('ZfcRbac\View\Strategy\RedirectStrategy'))->attach($eventManager);
```
same for others

svycka: this works for both zf2 and zf3

divix1988: Thanks @svycka I will test it today and update if that works in same way.

divix1988: You are correct, my bad.

divix1988: Done

divix1988: That's done now.

PR #368, opened 2017-04-24
Allow new zend-config:^3.1

svycka:
as this is new addition I like to add latest version and not a ^3.0 because 3.1 has more features and fixed some bugs


prolic:
thx :+1:

Issue #373, opened 2017-05-04
Cannot get a name and parameters for RoleProviderPluginManager


ashatrov:
Hi! I have fatal error with PHP7.0.18 + ZF2.5.3 + ZF-Commons/zfc-rbac 2.6.2

```
Fatal error: Uncaught Zend\ServiceManager\Exception\ServiceNotFoundException: ZfcRbac\Role\RoleProviderPluginManager::get was unable to fetch or create an instance for in /var/www/vhosts/pf-portal/vendor/zendframework/zend-servicemanager/src/ServiceManager.php:555
Stack trace:
#0 /var/www/vhosts/pf-portal/vendor/zendframework/zend-servicemanager/src/AbstractPluginManager.php(161): Zend\ServiceManager\ServiceManager->get(NULL, true)
#1 /var/www/vhosts/pf-portal/vendor/zf-commons/zfc-rbac/src/ZfcRbac/Factory/RoleServiceFactory.php(60): Zend\ServiceManager\AbstractPluginManager->get(NULL, false)
#2 /var/www/vhosts/pf-portal/vendor/zf-commons/zfc-rbac/src/ZfcRbac/Factory/RoleServiceFactory.php(77): ZfcRbac\Factory\RoleServiceFactory->__invoke(Object(Zend\ServiceManager\ServiceManager), 'ZfcRbac\\Service...')
#3 [internal function]: ZfcRbac\Factory\RoleServiceFactory->createService(Object(Zend\ServiceManager\ServiceManager), 'zfcrbacservicer...', 'ZfcRbac\\Service...')
#4 /var/www/vhosts/pf-portal/vendor/zendframework/zend-se in /var/www/vhosts/pf-portal/vendor/zendframework/zend-servicemanager/src/ServiceManager.php on line 555
```

The issue in \ZfcRbac\Factory\RoleServiceFactory

with this code:
```
$roleProvider = $pluginManager->get(key($roleProviderConfig), current($roleProviderConfig));
```

The internal pointer in array $roleProviderConfig is in the end of array, because of that `key($roleProviderConfig)` and `current($roleProviderConfig)` return null.

There is two options to fix it on place:
1. Reset internal pointer in array to the beginning:
```
reset($roleProviderConfig);
$roleProvider = $pluginManager->get(key($roleProviderConfig), current($roleProviderConfig));
```
2. Use different functions, which work without pointer.
```
[...]

PR #374, opened 2017-05-04
Cannot get a name and parameters for RoleProviderPluginManager #373

ashatrov:
https://github.com/ZF-Commons/zfc-rbac/issues/373

### DIFF
--- a/src/ZfcRbac/Factory/RoleServiceFactory.php
+++ b/src/ZfcRbac/Factory/RoleServiceFactory.php
@@ -56,6 +56,7 @@ class RoleServiceFactory implements FactoryInterface
         $pluginManager = $container->get('ZfcRbac\Role\RoleProviderPluginManager');
 
         /* @var \ZfcRbac\Role\RoleProviderInterface $roleProvider */
+        reset($roleProviderConfig);
         $roleProvider = $pluginManager->get(key($roleProviderConfig), current($roleProviderConfig));
 
         /* @var \Rbac\Traversal\Strategy\TraversalStrategyInterface $traversalStrategy */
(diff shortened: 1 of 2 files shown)


prolic:
Please add a test case that fails without this addition.

ashatrov-propertyfinder:
@prolic I have no idea how to simulate it.

Anyway, it is the correct way to get firsts key of Array, because the internal pointer can point beyond the end of the elements list.

http://php.net/manual/en/function.key.php
```
Return Values

The key() function simply returns the key of the array element that's currently being pointed to by the internal pointer. It does not move the pointer in any way. If the internal pointer points beyond the end of the elements list or the array is empty, key() returns NULL.
```

If we need to have the first key, it must be with reset(), because we cannot know where is pointer now.

ashatrov:
@prolic I have no idea how to simulate it.

Anyway, it is the correct way to get firsts key of Array, because the internal pointer can point beyond the end of the elements list.

http://php.net/manual/en/function.key.php
```
Return Values

The key() function simply returns the key of the array element that's currently being pointed to by the internal pointer. It does not move the pointer in any way. If the internal pointer points beyond the end of the elements list or the array is empty, key() returns NULL.
```

If we need to have the first key, it must be with reset(), because we cannot know where is pointer now.

ashatrov:
e.g.

```
<?php
$arr = ['foo' => 1];

echo key($arr); // foo

next($arr);
echo key($arr); // Null

reset($arr);
echo key($arr); // foo
```

Maybe there is `while(next($config)) { // do something }` somewhere in ZF.

humayunjamal:
Hi Guys waiting for this pr to be merged

ashatrov-propertyfinder:
@prolic What the policy for releases? Will you create a minor release 2.6.3 with this changes to be able to use it in composer from packagist?

prolic:
@ashatrov-propertyfinder done with https://github.com/ZF-Commons/zfc-rbac/releases/tag/v2.6.3

PR #366, opened 2017-04-05
Develop

basz:
- switch to php7
- add return typehint
- switch phpunit 6
- include zfr-rbac in library
- adds header check
- switch travis notifications

PR #379, opened 2018-02-18
Multiple assertions develop

basz:
Implements https://github.com/ZF-Commons/zfc-rbac/pull/320 for develop branch

- does so without performance loss if not used
- assertion sets may be nested with different conditions
- does so lazily for each condition
  - an AND condition will break early as soon as one assertions fails
  - an OR condition will break early as soon as one assertions succeed
- assertions are cached, so a next call won't require to construct an assertion or AssertionSet.

documentation forthcoming


basz:
no sure why this PR also show commits already pulled into develop.

review request is from commit e7fd215 and forward

src/Role/HierarchicalRole.php
```diff
@@ -0,0 +1,47 @@
...
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license.
+ */
+
+namespace ZfcRbac\Role;
+
+/**
+ * Simple implementation for a hierarchical role
+ */
+final class HierarchicalRole extends Role implements HierarchicalRoleInterface
```
svycka: maybe if we use `final` then add to others also?

DavidHavl:
Looks great!!

basz:
@svycka thanks. yes, I will do additional changes like php71, remove zf2 support etc... for now could you have a look at the changes from commit e7fd215 and forward?

svycka:
> @svycka thanks. yes, I will do additional changes like php71, remove zf2 support etc... for now could you have a look at the changes from commit e7fd215 and forward?

@basz I already did, but if you could rebase would be a lot easier to review because now it is very hard to see relevant changes.

basz:
I now figured out what goed wrong. this PR https://github.com/ZF-Commons/zfc-rbac/pull/366 isn't merged into develop yet. Guess I should fix that one first and then rebase this one.

svycka:
@basz  or you can remove commits before e7fd215 also should work

composer.json
```diff
@@ -27,15 +27,15 @@
...
+    "php": "^7.0",
+    "zendframework/zend-servicemanager": "^3.0.3",
+    "zendframework/zend-stdlib": "^3.0"
   },
   "require-dev": {
-    "phpunit/phpunit": "~5.0",
-    "squizlabs/php_codesniffer": "^2.0",
-    "doctrine/common": "~2.4"
+    "malukenho/docheader": "^0.1",
+    "phpunit/phpunit": "^6.0",
```
prolic: let's use v7

src/Assertion/AssertionSet.php
```diff
@@ -0,0 +1,119 @@
...
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license.
+ */
+
+namespace ZfcRbac\Assertion;
+
+use ZfcRbac\Exception;
+use ZfcRbac\Identity\IdentityInterface;
+
+class AssertionSet implements AssertionInterface
```
prolic: if it's implementing an interface, mark as final

src/Role/HierarchicalRole.php
prolic: agree, when it's implementing an interface and has no additional public methods, let's mark as final

src/Role/Role.php
```diff
@@ -1,4 +1,6 @@
 <?php
```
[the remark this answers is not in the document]
prolic: I remember there is some git plugin that fixes that, but I forgot the name.

src/Role/Role.php
```diff
@@ -16,68 +18,52 @@
...
+namespace ZfcRbac\Role;
 
 /**
- * @ORM\Entity
- * @ORM\Table(name="permissions")
+ * Simple implementation for a role without hierarchy
+ * and using strings as permissions
  */
-class Permission implements PermissionInterface
+class Role implements RoleInterface
```
prolic: same as above, if it's implementing an interface, mark as final

test/Assertion/AssertionSetTest.php
```diff
@@ -0,0 +1,262 @@
...
+use ZfcRbac\Exception\InvalidArgumentException;
+use ZfcRbac\Identity\IdentityInterface;
+use ZfcRbacTest\Asset\SimpleAssertion;
+
+/**
+ * @covers \ZfcRbac\Assertion\AssertionSet
+ */
+class AssertionSetTest extends TestCase
+{
+    public function testImplementsAssertionInterface()
```
prolic: do we require php 7.1 now? if so, let's add `: void`

test/Asset/MockRoleWithPermissionMethod.php
```diff
@@ -11,12 +29,13 @@ public function getPermissions()
...
-    public function getName()
+    public function getName(): string
     {
         return 'role-with-permission-method';
     }
-    public function hasPermission($permission)
+
+    public function hasPermission(string $permission): bool
     {
         return false;
```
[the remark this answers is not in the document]
prolic: If someone has a problem when using a MockRole from TestAsset in production, he deserves this problem :)

prolic:
Looks good so far, just some minor notes

PR #380, opened 2018-02-21
Plugin managers refactor and more improvements

svycka:
this way we could make another step to decouple from Zend service-manager and fully support any PSR-11 implementation. If everyone agrees I will continue with `RoleProviderPluginManager` maybe more don't remember if we have more :D

I guess would be a good idea to change config
````php
ZfcRbac\Assertion\AssertionPluginManager::class => ZfcRbac\Container\AssertionPluginManagerFactory::class,
// to
ZfcRbac\Assertion\AssertionContainerInterface::class => ZfcRbac\Container\AssertionPluginManagerFactory::class,
````

Also not sure about naming maybe `ZfcRbac\Assertion\ContainerInterface` or `ZfcRbac\Container\AssertionContainerInterface`

what do you think? @prolic @basz

### CODE REVIEW

File: src/Assertion/AssertionPluginManager.php [hunk 1]
```diff
@@ -33,8 +33,11 @@
 {
     protected $instanceOf = AssertionInterface::class;
 
-    public function get($name): AssertionInterface
+    /**
+     * {@inheritdoc}
+     */
```
basz: personally, I don't see the point of such API docs. Especially in modern IDE's

svycka: I added this because of IDE :D without this it complains about uncaught exceptions

src/Assertion/AssertionContainerInterface.php
```diff
@@ -19,29 +19,9 @@
...
-/**
- * Plugin manager to create role providers
- *
- * @method RoleProviderInterface get($name)
- *
- * @author  Michaël Gallego <mic.gallego@gmail.com>
- * @license MIT
- */
-final class RoleProviderPluginManager extends AbstractPluginManager
+interface AssertionContainerInterface
```
danizord: What about extending PSR-11 `ContainerInterface` here?

src/Assertion/AssertionContainerInterface.php
```diff
@@ -19,29 +19,9 @@
...
-    protected $instanceOf = RoleProviderInterface::class;
-
-    /**
-     * @var array
-     */
-    protected $factories = [
-        InMemoryRoleProvider::class         => InvokableFactory::class,
-        ObjectRepositoryRoleProvider::class => ObjectRepositoryRoleProviderFactory::class,
-    ];
+    public function get($name): AssertionInterface;
```
prolic: typehint string

test/Container/RoleServiceFactoryTest.php
```diff
@@ -44,10 +44,10 @@ public function testCanCreateRoleService(): void
             ],
         ]);
 
-        $container = $this->getMockBuilder(ContainerInterface::class)->getMock();
-
-        $container->expects($this->at(0))->method('get')->with(ModuleOptions::class)->willReturn($options);
-        $container->expects($this->at(1))->method('get')->with(RoleProviderPluginManager::class)->willReturn(new RoleProviderPluginManager($this->getMockBuilder(ContainerInterface::class)->getMock()));
+        $container = new ServiceManager(['services' => [
```
prolic: if we go psr-11, let's remove service-manager in tests, too and use prophecy for mocking the interface

src/Assertion/AssertionContainerInterface.php
svycka: I can't because of `AbstractPluginManager::get($name, array $options = null)`

src/Assertion/AssertionContainerInterface.php
svycka: Not sure about this. What would we get from this? but yes maybe we can.
@prolic @basz what you think?

test/Container/RoleServiceFactoryTest.php
svycka: this PR is not about removing service manager but it brings an ability to use something else if needed. If we would decide to remove it we can do it in separate PR I guess.

src/Container/ObjectRepositoryRoleProviderFactory.php
```diff
@@ -33,8 +33,10 @@
  */
 final class ObjectRepositoryRoleProviderFactory
 {
-    public function __invoke(ContainerInterface $container, $requestedName, ?array $options = []): ObjectRepositoryRoleProvider
+    public function __invoke(ContainerInterface $container): ObjectRepositoryRoleProvider
     {
+        $options = $container->get('config')['zfc_rbac']['role_provider'][ObjectRepositoryRoleProvider::class] ?? [];
```
svycka: I am not sure about `$container->get('config')` this is ZF thing. Maybe someone can suggest something?

test/Container/RoleServiceFactoryTest.php
basz: Is this article also true for phpunit? http://everzet.com/post/72910908762/conceptual-difference-between-mockery-and-prophecy. I'm only familiar with phpunit, perhaps I should broaden my horizon

src/Container/ObjectRepositoryRoleProviderFactory.php
basz: how do Symfony and others pull config from their container? Looks pretty standard to me, dunno if we could generalize some this trivial...

src/Assertion/AssertionContainerInterface.php
basz: why not. it's a container

src/Container/ObjectRepositoryRoleProviderFactory.php
svycka: I think I could use `ZfcRbac\Options\ModuleOptions` here will try but anyway `ZfcRbac\Options\ModuleOptions` itself uses `$container->get('config')['zfc_rbac']` also maybe we don't need those factories? It's not that hard for the user to write themselves the way they need and want. This would also allow us to remove role_provider config option is a good thing no?

test/Container/RoleServiceFactoryTest.php
svycka: @prolic  I can mock this if you think this is really needed

config/dependencies.global.php
```diff
@@ -23,7 +23,8 @@
         'factories' => [
             ZfcRbac\Assertion\AssertionPluginManager::class      => ZfcRbac\Container\AssertionPluginManagerFactory::class,
             ZfcRbac\Options\ModuleOptions::class                 => ZfcRbac\Container\ModuleOptionsFactory::class,
-            ZfcRbac\Role\RoleProviderPluginManager::class        => ZfcRbac\Container\RoleProviderPluginManagerFactory::class,
+            ZfcRbac\Role\InMemoryRoleProvider::class             => ZfcRbac\Container\InMemoryRoleProviderFactory::class,
+            ZfcRbac\Role\ObjectRepositoryRoleProvider::class     => ZfcRbac\Container\ObjectRepositoryRoleProviderFactory::class,
```
svycka: can we vote for removing those factories?
@prolic @basz

src/Service/RoleService.php
```diff
@@ -37,33 +37,19 @@
...
-    {
-        $this->roleProvider = $roleProvider;
-    }
-
-    public function setGuestRole(string $guestRole): void
-    {
         $this->guestRole = $guestRole;
     }
 
-    public function getGuestRole(): string
```
svycka: not sure about this one but why we need this? If we need for some reason maybe we can get from `ZfcRbac\Options\ModuleOptions`

basz: agree, remove

src/Container/ObjectRepositoryRoleProviderFactory.php
```diff
@@ -33,8 +34,11 @@
  */
 final class ObjectRepositoryRoleProviderFactory
```
svycka: maybe remove this factory?
@prolic @basz

src/Container/InMemoryRoleProviderFactory.php
```diff
@@ -22,20 +22,22 @@
...
 /**
- * Factory to create a role provider plugin manager
+ * Factory used to create an in memory role provider
  *
- * @author  Michaël Gallego <mic.gallego@gmail.com>
+ * @author  Vytautas Stankus
  * @licence MIT
  */
-final class RoleProviderPluginManagerFactory
+final class InMemoryRoleProviderFactory
```
svycka: maybe remove this factory?
@prolic @basz

src/Assertion/AssertionContainer.php
```diff
@@ -28,10 +28,16 @@
...
+final class AssertionContainer extends AbstractPluginManager implements AssertionContainerInterface
 {
     protected $instanceOf = AssertionInterface::class;
+
+    /**
+     * {@inheritdoc}
+     */
+    public function get($name, array $options = null): AssertionInterface
+    {
+        return parent::get($name, $options);
```
svycka: Maybe better go safely here and don't pass $options here? since we don't have this in an interface

svycka:
I think it's ready for review and I am waiting for answers to my questions

src/Assertion/AssertionContainer.php
svycka: another solution would be we could extend ServiceManager instead of PluginManager

src/Container/ObjectRepositoryRoleProviderFactory.php
svycka: for me, it's not that hard to create a factory so I would say remove. And maybe if they have config somewhere else or something they anyway have to create it. But for most people, this is enough so don't know remove or no.

basz:
Lets keep it then

svycka:
I have added tests and if we decide to keep 'role_provider' config then its ready to merge unless someone sees something bad.

basz:
great work, so much removed!

PR #381, opened 2018-02-22
covers role with no 'children'

basz:
covers

https://coveralls.io/builds/7406433/source?filename=src%2FRole%2FInMemoryRoleProvider.php#L81


svycka:
if you working with coverage maybe fix coveralls setup somehow does not work with develop branch

PR #382, opened 2018-02-22
Refactored AuthorizationService

svycka:
how about this?

### DIFF
--- a/src/Service/AuthorizationService.php
+++ b/src/Service/AuthorizationService.php
@@ -80,7 +80,13 @@ final class AuthorizationService implements AuthorizationServiceInterface
         }
 
         if (! empty($this->assertions[$permission])) {
-            $assertionSet = new AssertionSet($this->assertionPluginManager, $this->assertions);
+            if (! \is_array($this->assertions[$permission])) {
+                $permissionAssertions = [$this->assertions[$permission]];
+            } else {
+                $permissionAssertions = $this->assertions[$permission];
+            }
+
+            $assertionSet = new AssertionSet($this->assertionPluginManager, $permissionAssertions);
 
             return $assertionSet->assert($permission, $identity, $context);
         }
--- a/src/Service/AuthorizationService.php
+++ b/src/Service/AuthorizationService.php
@@ -46,14 +46,14 @@ final class AuthorizationService implements AuthorizationServiceInterface
     private $roleService;
 
     /**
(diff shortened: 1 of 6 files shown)


src/Service/AuthorizationService.php
```diff
@@ -98,45 +79,12 @@ public function isGranted(?IdentityInterface $identity, string $permission, $con
             return false;
         }
 
-        if ($this->hasAssertion($permission)) {
-            return $this->assert($this->assertions[$permission], $permission, $identity, $context);
-        }
+        if (! empty($this->assertions[$permission])) {
+            $assertionSet = new AssertionSet($this->assertionPluginManager, $this->assertions);
```
svycka: yes, we will have to create one more object but is it that bad? But this will allow us to create nested assertions example:
````php
[
    'zfc_rbac' => [
        'assertion_map' => [
            // single assertion
            'myPermission' => 'myAssertion',
            'myPermission2' => [
                // multiple assertions
                'isAdmin',
                [
                    'isSharedWithEditPermission',
                    'isOwner',
                    // condition
                    'condition' => \ZfcRbac\Assertion\AssertionSet::CONDITION_AND,
                ],
                // condition
                'condition' => \ZfcRbac\Assertion\AssertionSet::CONDITION_OR,
            ],
        ],
    ],
];
````
at least I think this should work

Comment (basz) [reply src/Service/AuthorizationService.php]: Nested assertions are already possible.

I noticed roles are checked first and when a particular role is allowed and assertions are defined these are evaluated. So in this example `isAdmin` really makes now sense (that is a role?).

File: src/Service/AuthorizationService.php [hunk 2]
```diff
@@ -98,45 +79,12 @@ public function isGranted(?IdentityInterface $identity, string $permission, $con
...
-            $this->assertions[$permission] = $assertion = new AssertionSet($this->assertionPluginManager, $assertion);
-
-            return $assertion->assert($permission, $identity, $context);
-        }
-
-        throw new Exception\InvalidArgumentException(sprintf(
-            'Assertion must be callable, string, array or implement ZfcRbac\Assertion\AssertionInterface, "%s" given',
-            is_object($assertion) ? get_class($assertion) : gettype($assertion)
-        ));
+        return true;
```
basz: To move towards AssertionSet here is good

svycka:
I think it's ready to merge

Issue #387, opened 2019-02-18
HierarchicalRoleInterface problem


svycka:
here https://github.com/ZF-Commons/zfc-rbac/blob/develop/src/Role/HierarchicalRoleInterface.php#L36 maybe we could change the return type to `iterable`?
````php
public function getChildren(): array;
// to
public function getChildren(): iterable;
````
Then when working with doctrine I could just return ArrayCollection, but now I have to do this:
````php
public function getChildren(): array
{
    return $this->children->toArray();
}
`````

basz:
Sounds reasonable

PR #388, opened 2019-02-19
changed types from array to iterable and other improvements

svycka:
closes https://github.com/ZF-Commons/zfc-rbac/issues/387

changed not only in interface but maybe better be consistent?


svycka:
ping @prolic

PR #394, opened 2020-01-16
Minor tweak

basz:
- Test php73+ support
- add 2 return types


basz:
should we drop php71 support? phpunit85 does not work with 71

PR #393, opened 2020-01-15
Laminas migration (development)

basz:

PR #395, opened 2020-05-07
Added config option for authorization service instead of hard-coded one

svycka:
There are problems with migration to laminas I can't change `Zend\Authentication\AuthenticationService` to `Laminas\Authentication\AuthenticationService`
then I add an alias in service manager for this but in some cases, I can't use this workaround and would be nice if this would be not hardcoded.
now you can just add to config this
```php
'zfc_rbac' => [
  'authentication_service' => 'Laminas\Authentication\AuthenticationService' // or 'zfcuser_auth_service' in case of zfcUser
];
```
This is not a BC break because default still is `Zend\Authentication\AuthenticationService`

### DIFF
--- a/docs/07. Cookbook.md
+++ b/docs/07. Cookbook.md
@@ -750,10 +750,8 @@ To use the authentication service from ZfcUser, just add the following alias in 
 
 ```php
 return [
-    'service_manager' => [
-        'aliases' => [
-            'Zend\Authentication\AuthenticationService' => 'zfcuser_auth_service'
-        ]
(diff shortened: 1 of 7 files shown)


prolic:
we probably should migrate to laminas and get rid of zend* deps. @basz, thoughts?

svycka:
@prolic even if you migrate, this would still be useful don't you think? Anyway, having this hardcoded was bad idea in the first place.

basz:
I agree with @svycka . this should be merged regardless of the state of #392 or #393

